PR #13, opened 2022-01-20
Feature/41 split repo

albertcrowley:
* Remove front end directories
* Removed references to API in Circle CI configuration
* Simplified the CI jobs names
* Updated local CI run instructions
* Adding an early check in CI script for required environment variables


mjtravers (approved):
Both Docker and CircleCI config files look good. I was able to run the docker-compose set up locally without any issues.

PR #16, opened 2022-01-26
Removed 'key' and 'api' from git secrets scan rules

mjtravers:
Changes made to git secrets install script per 1/26/2022 meeting discussion.


lbeaufort (approved):
Looks great, thanks for making this change so quickly!

PR #17, opened 2022-01-27
Deploy to cloud.gov

lbeaufort:
## Summary of changes
- Made sure naming conventions for apps and services make sense/are consistent across apps (documented here: https://docs.google.com/spreadsheets/d/1P-tA0xVnlVDya4TLj5hkRxyenBHl7ceLbrIUI_HmNus/edit#gid=0)
-  Made production settings file with `debug=False` (and possibly `CORS_ALLOWED_ORIGINS`)
- Removed `DJANGO_SECRET_KEY` from settings
- Set `DJANGO_SECRET_KEY` in cloud.gov dev/stage/prod envs. Put in follow-up tickets to rotate.
- Restored full database backup to dev/stage/prod
- Hooked up application and database with [dj-database-url](https://pypi.org/project/dj-database-url/) (uses built-in `$DATABASE_URL` - see https://docs.cloudfoundry.org/devguide/deploy-apps/environment-variable.html#DATABASE-URL
- Moved requirements.txt and requirements-test.txt to root directory
- Added `wait_for_db` in deploy command and made sure it works
- Made deploy task
- Added deploy task to CircleCI 
- Added service keys for dev/stage/prod, CF username/password env vars to Circle for dev/stage/prod (see README)
- Added `$DATABASE_URL` env var to CircleCI
- Dockerfile changes
- Renamed `FECFILE_DB_NAME` to `FECFILE_TEST_DB_NAME`
- Made sure deploy tasks are in alignment in web app and api (https://github.com/fecgov/fecfile-web-app/pull/85/files)
- General code cleanup

## Steps to test (QA)

**Testing URLS**
https://fecfile-web-api-dev.app.cloud.gov/api/v1/token/obtain
https://fecfile-web-api-stage.app.cloud.gov/api/v1/token/obtain
https://fecfile-web-api-prod.app.cloud.gov/api/v1/token/obtain

## Steps to test/additional developer information (devs)
**Deploy with invoke**
- Log in to cloud.gov with `cf login --sso`
- `invoke deploy --space dev`

**Deploy with CircleCI**
- Make a copy of this branch
- Change deploy rules to deploy your branch
- Commit/push to github

**Locally**
- Set `DATABASE_URL`?
- Run with Docker

**(Optional) Create RDS**
- ` cf create-service aws-rds small-psql-redundant fecfile-api-rds`
- Wait for db creation
- Bind to app in manifest, redeploy

**(Optional) Restore database from backup**
- Get database dump files from Google drive, see Matt for password
- See https://cloud.gov/docs/services/relational-database/ and https://github.com/fecgov/fec-cms/wiki/Backup-or-Restore-Wagtail-PostgreSQL-database-in-dev,-staging,-or-production but I used the service connect plugin
- `cf connect-to-service -no-client fecfile-web-api fecfile-api-rds`
- New tab
- `psql --dbname (from ssh window) --host localhost --port (from ssh window)  --username (from ssh window)   < fec_clean_dev_db_backup-20211227.sql`

**Testing URLS**
https://fecfile-web-api-dev.app.cloud.gov/api/v1/token/obtain
https://fecfile-web-api-stage.app.cloud.gov/api/v1/token/obtain
https://fecfile-web-api-prod.app.cloud.gov/api/v1/token/obtain


.cfignore
```diff
@@ -0,0 +1,2 @@
+# Database creation script
+db/
```
lbeaufort, line 2: double-check this

.circleci/config.yml
```diff
@@ -13,14 +13,15 @@ jobs:
 
   dependency-check:
     docker:
-      - image: cimg/python:3.8
+      - image: cimg/python:3.7
```
albertcrowley, line 17: It looks like we are using python 3.8 in other configs - but I just logged into the running fecfile-web-api container running on cloud.gov and python3 is version Python 3.6.9 - but if you just run "python" and not "python3" you get Python 2.7.17.   Setting the Python 2/3 that aside for now, this makes me think that we should standardize everything (dev docker containers and CircleCI images) on Python 3.6.9 if that is what is really running on cloud.gov.  

Back to the Python 2 thing - Looking at our scripts, we don't specify python3, but I think it won't run on python2 at all.  I'm thinking that something is causing our app to run in python 3 on cloud.gov but I don't see where it happens. Laura, if you are doing that explicitly somewhere, can you let me know how for educational purposes?

.circleci/config.yml
```diff
@@ -68,29 +69,28 @@ jobs:
...
           working_directory: ~/project/django-backend/
 
       - run:
           name: Load test database fixure
           command: |
             sudo apt-get update &&
             sudo apt-get install postgresql-client-12 &&
-            export PGPASSWORD=${FECFILE_DB_PASSWORD} &&
-            psql -h ${FECFILE_DB_HOST}  ${FECFILE_DB_NAME} ${FECFILE_DB_USERNAME} < fec_clean_dev_db_backup-20211227.sql
+            psql ${DATABASE_URL} < fec_clean_dev_db_backup-20211227.sql
```
albertcrowley, line 93: I didn't know you could pass the connection string to psql.  Nice.

django-backend/fecfiler/settings/production.py
```diff
@@ -0,0 +1,18 @@
+from .base import *  # noqa F401, F403
```
albertcrowley, line 1: This is still getting flagged in SonarCloud. I think their decorator to skip linting, and all other static checks on the line, is "# NOSONAR"
Not sure, but maybe you can stack the noqa with the NOSONAR like:
`# NOSONAR noqa F401, F403`

django-backend/fecfiler/settings/base.py
```diff
@@ -1,19 +1,15 @@
...
 import datetime
+import dj_database_url
+
+from .env import env
 from corsheaders.defaults import default_headers
-import logging
+from django.utils.crypto import get_random_string
 
 
 CORS_ALLOWED_ORIGINS = [
```
albertcrowley, line 15: When I access the cloud.gov [front end](https://fecfile-web-app-dev.app.cloud.gov/#/) and login, I'm seeing some CORS errors in the console.  It's an odd thing because some API responses do have the proper `access-control-allow-origin: *
` header and others don't.  For example https://fecfile-web-api-dev.app.cloud.gov/api/v1/user/login/authenticate does have the header and so login works fine.     

The ones that don't have the CORS header generate errors in the browser dev console:
`Cross-Origin Request Blocked: The Same Origin Policy disallows reading the remote resource at https://fecfile-web-api-dev.app.cloud.gov/api/v1/core/recent_saved_reports. (Reason: CORS header ‘Access-Control-Allow-Origin’ missing). Status code: 502.`
It seems strange that some are getting the header and others don't.  If this is an simple fix, it is in scope to do it in this ticket and it makes sense to get it right - but if it is a bigger problem or a hard fix, I would support rolling it into a new issue.

albertcrowley (changes requested):
I put inline comments in the PR.  The summary is:
* I have a question about the Python version we should be using in general for our CI and dev containers.
* I'm seeing CORS errors in the web console when talking to the cloud.gov API so we need to either fix that up or spin it out to a new issue for later.

django-backend/fecfiler/settings/base.py
lbeaufort: Thanks for flagging this, it looks like the `gunicorn` workers were rebooting due to high CPU, resulting in 502/503 errors without CORS headers. I redeployed with more memory, which looks like that took care of it. Can you take a look and see if it's working properly now?

.circleci/config.yml
lbeaufort: That's interesting! It looks like:
>The difference when you ssh in is that the profile.d scripts which setup PATH etc... are not automatically run (again, for the app runtime they are). To obtain the same environment as runtime, please run /tmp/lifecycle/shell when you first SSH in.
ref: https://github.com/cloudfoundry/python-buildpack/issues/98#issuecomment-366244014

django-backend/fecfiler/settings/base.py
albertcrowley: I just tested at https://fecfile-web-app-dev.app.cloud.gov/ and all my CORS errors are gone now.  So this looks like a fix to me.

.circleci/config.yml
lbeaufort: It looks like so far (if we go off the Dockerfile) this app is running on 3.7, and it's probably best to keep Circle in alignment with the app is running. Since this PR is already quite big, I think it might be easier to put in a separate issue to update the python version to 3.8.

lbeaufort: Also, using `cf ssh fecfile-web-api` and `/tmp/lifecycle/shell`, I confirmed that cloud foundry respected `runtime.txt` and is running Python 3.7.12.

albertcrowley (approved):

django-backend/fecfiler/settings/production.py
lbeaufort: Fixed, thank you!

mjtravers (approved):
That's a lot of stuff. Good code clean up, too.

PR #28, opened 2022-02-01
Removed unused legacy files

mjtravers:
The buildspec.yml and cf-setup.sh files are no longer in use.


albertcrowley (approved):
I agree, these files are now OBE and can be removed.

PR #41, opened 2022-02-09
Update postgres connection string in README

lbeaufort:


lbeaufort:
Simple README change, so merging.

PR #43, opened 2022-02-09
Removed gunicorn `reload` flag in production

lbeaufort:
Remove `reload` flag in production, Per https://docs.gunicorn.org/en/stable/settings.html#reload
>This setting is intended for development. It will cause workers to be restarted whenever application code changes.


albertcrowley (commented):
This looks correct to me.  We don't need the --reload flag when deployed on cloud.gov and that is the only time that the app is started using the run.sh script.

PR #53, opened 2022-02-18
Feature/34 contacts model

toddlees:


django-backend/fecfiler/contacts/migrations/0001_initial.py
```diff
@@ -0,0 +1,77 @@
...
+                ),
+                ("candidate_id", models.TextField(blank=True, null=True)),
+                ("committee_id", models.TextField(blank=True, null=True)),
+                ("name", models.TextField(blank=True, null=True)),
+                ("last_name", models.TextField(blank=True, null=True)),
+                ("first_name", models.TextField(blank=True, null=True)),
+                ("middle_name", models.TextField(blank=True, null=True)),
+                ("prefix", models.TextField(blank=True, null=True)),
+                ("suffix", models.TextField(blank=True, null=True)),
+                ("street_1", models.TextField()),
```
mjtravers, line 46: Looking at all 4 contact specs, the fields street_1, city, state, zip, and country are required for all of them. Let's remove the null=True for those fields in the model so that nulls can't be entered.

mjtravers (changes requested):
I see the deleted is a timestamp rather than boolean, which is good. We can still query for non-deleted records be we also get the date the record was deleted so good change. I'll update the acceptance criteria.

PR #63, opened 2022-03-03
Added missing migrations, removed `makemigrations` commands from deployment

lbeaufort:
Issue #67 

**Changes**
- Add missing migrations
- Tests can now fail when we forget to check in migrations. We were going to always miss this this because Circle was running `makemigrations` for us, which is pretty risky
- In addition, the `bin/run.sh` command for deployment to cloud.gov will no longer make migration. The migrations should already be available in the code.

**Background/next steps**
- Going forward, migration files should be generated by the developers and checked in with changes. The best way to do may be for the developer to complete code changes,  log into the docker image (`docker exec -it fecfile-api /bin/bash`), run `./manage.py makemigrations`, and commit the migration(s) along with code changes.
- One of the risks of not checking in these missing initial migrations is that there could eventually be a fairly complex migration conflicts.
- We should keep synchronization in mind going forward to make sure nothing behaves unexpectedly

**How to test**
Check out this branch, run the app in docker,  run tests by opening a new tab and logging into the docker image (`docker exec -it fecfile-api /bin/bash`) and running `./manage.py migrate` and `./manage.py test --keepdb`. Previously needed to make migrations for tests to pass.


toddlees (approved):
Reviewed while it was written

PR #84, opened 2022-03-23
Feature/29 fix api docs remove nested router

toddlees:
resolves conflicts from backend reorg and removes nested router


lbeaufort (approved):
Thanks @toddlees!

Issue #78, opened 2022-03-18
Do not let the "deleted" column of models get updated by users


mjtravers:
The deleted column of model database tables records the timestamp when a REST delete request is made to the api.
Currently, a PUT or POST request to the api can update or insert a value into the deleted column when only the api system should be able to affect the field.

## Acceptance Criteria ##
- A PUT or POST request to a model table cannot make a change or insert a value into the "deleted" timestamp column

### QA ###
Unit test verification of delete feature: https://app.circleci.com/pipelines/github/fecgov/fecfile-web-api/261/workflows/28f5f70d-28aa-4a68-ae03-d23b1d7158a8/jobs/660

WiseQA:
Verified unit test passed successfully as denoted in the Acceptance Criteria.
![image.png](https://images.zenhubusercontent.com/61ba01e428a658b4eb0ca758/7383f47b-7f5f-4f5f-8828-678af0e84d32)

Verified -  Attempt to POST a new contact and set deleted to a timestamp and Attempt to PUT to (update) an existing contact an set it's deleted field (see above screenshots)

QA review completed.  Moved to Stage Ready.

PR #85, opened 2022-03-24
[Merge after release cut] makes delete readonly on contact serializer

toddlees:
## Summary

- Issue number #78 

This prevents PUTs and POSTs to our API from writing to the deleted field which would cause unexpected behavior

## How to test

Attempt to POST a new contact and set `deleted` to a timestamp
image
You will find that `deleted` has not been written to.  The fact that the contact comes back is further proof that `deleted` was not written to.

Attempt to PUT to (update) an existing contact an set it's `deleted` field
image
You will find that `deleted` has not been written to.  The fact that the contact comes back is further proof that `deleted` was not written to.

-

### DIFF
--- a/django-backend/fecfiler/contacts/serializers.py
+++ b/django-backend/fecfiler/contacts/serializers.py
@@ -15,10 +15,10 @@ class ContactSerializer(serializers.ModelSerializer):
         """
 
         contact_value = dict(
-            COM='Committee',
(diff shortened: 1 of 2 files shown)


mjtravers (approved):
That's a ridiculously small amount of code to review for this feature. Nice job.

PR #92, opened 2022-03-30
Created F3X API view and integrate validation

toddlees:
#7 
Adds F3X crud endpoints and integrates fecfile-validate into the new serializer.
Removes staticfiles
Cleans up log settings and adds more logging to apps

To test you can 
POST to `/api/v1/f3x-summaries/ `with a valid f3x in json, like 
```
"form_type": "F3XN",
"filer_committee_id_number": "C00000000",
"treasurer_last_name": "Lastname",
"treasurer_first_name": "Firstname",
"date_signed": "20040729",
"created": "2022-02-09T00:00:00.000Z",
"updated": "2022-02-09T00:00:00.000Z"
```
PUT to `/api/v1/f3x-summaries/<id>/` with valid f3x in json to update
DELETE to `/api/v1/f3x-summaries/<id>/` to soft delete
GET to `/api/v1/f3x-summaries/<id>/` to get an f3x

see validation errors (note the 400 response):
image


mjtravers (changes requested):
SonarCloud is reporting 2 very minor code smells. It looks like it wants "%s" rather than {} in the logging string interpolation.

> recommended import logging
> logging.error("Error: User %s has not been able to access %s", "Alice", "MyFile")

Sending back for code smell fixes

django-backend/fecfiler/f3x_summaries/serializers.py
```diff
@@ -0,0 +1,168 @@
...
+            translated_errors = reduce(collect_error, validation_result.errors, {})
+            logger.warning(
+                f"F3X Summary: Failed validation for {list(translated_errors)}"
+            )
+            raise exceptions.ValidationError(translated_errors)
+        return data
+
+    class Meta:
+        model = F3XSummary
+        fields = [
```
mjtravers, line 35: Not a priority but just wondering: Is there a way for the serializer to get the fields from the model properties so we don't have to keep the 2 in sync? Or are there differences I am missing?

lbeaufort:
> %s",

@mjtravers @toddlees the `%s` is generally considered the "old" way of string formatting. Alternately, here we could use a `.format` or a f-string to satisfy SonarCloud.

django-backend/fecfiler/f3x_summaries/models.py
```diff
@@ -28,12 +22,12 @@ class F3XSummary(SoftDeleteModel):
     coverage_from_date = models.TextField(null=True, blank=True)
     coverage_through_date = models.TextField(null=True, blank=True)
     qualified_committee = models.BooleanField(default=False, null=True, blank=True)
-    treasurer_last_name = models.TextField(null=True, blank=False)
-    treasurer_first_name = models.TextField(null=True, blank=False)
+    treasurer_last_name = models.TextField(null=True, blank=True)
```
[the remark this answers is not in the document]
toddlees: i'm setting them all to blank=True so that the field validation does not short circuit the drf validation

django-backend/fecfiler/f3x_summaries/serializers.py
toddlees: the only difference is `deleted` is not included.  but if there's a way to get the list dynamically, we could dynamically remove `deleted` i would assume.  I'll look into it

toddlees:
> > SonarCloud is reporting 2 very minor code smells. It looks like it wants "%s" rather than {} in the logging string interpolation.
> > > recommended import logging
> > > logging.error("Error: User %s has not been able to access %s", "Alice", "MyFile")
> > 
> > 
> > Sending back for code smell fixes
> 
> @mjtravers @toddlees the `%s` syntax is generally considered the "old" way of string formatting. Alternately, here we could use a `.format()` or a f-string to satisfy SonarCloud.

@mjtravers @lbeaufort I've been using f-strings.  this was legacy code that must have gotten autoformatted when i changed the imports at the top. I can fix it though

PR #98, opened 2022-04-13
Release sprint-6 to production

mjtravers:


django-backend/fecfiler/committee_accounts/models.py
```diff
@@ -0,0 +1,34 @@
+from django.db import models
+from django.core.validators import RegexValidator
+from fecfiler.soft_delete.models import SoftDeleteModel
+
+COMMITTEE_ID_REGEX = RegexValidator(r"^C[0-9]{8}$", "invalid committee id format")
```
toddlees, line 5: This isn't necessarily the ultimate committee id pattern.  In the F3X schema it can start with P, H, or S as well.  We should probably pull the validation from the validation package.  I don't think we want much validation at the model-field level though.

toddlees (approved):

PR #105, opened 2022-04-28
Add DRF result ordering for f3x_summaries

mjtravers:
## Summary

- Issue number [app#165](https://app.zenhub.com/workspaces/fecfile-online-619e578e68408b001c831251/issues/fecgov/fecfile-web-app/165)

- Added QuerySet ordering to f3x_summaries
- Added alias() call in QuerySet to sort by report_code label
- Change committee account of contact test data to match that of the f3x_summaries test data (all go to QA's test account)


django-backend/fecfiler/f3x_summaries/views.py
```diff
@@ -13,6 +15,36 @@ class F3XSummaryViewSet(CommitteeOwnedViewSet):
     in CommitteeOwnedViewSet's implementation of get_queryset()
     """
 
-    queryset = F3XSummary.objects.all().order_by("-id")
+    queryset = F3XSummary.objects.alias(report_code_label=Case(
```
toddlees, line 18: along with what i suggested in the app PR, could this mapping live in a data structure that lives here and is passable to the front end?

mjtravers: If the data were dynamic and changed with some frequency I would want the labels to be pulled from one place. However, considering the stability of the labels and the rarity in which they will change (if ever) I think it better to forgo an HTTP call and maintain 2 separate lists.

toddlees (approved):

toddlees (approved):

PR #107, opened 2022-05-03
fixes validator headers

toddlees:
adds fields_to_validate to cors headers.  in the future we will rewrite this header to be a query parameter


mjtravers (approved):
Looks good. Will merge when deployment window opens

PR #108, opened 2022-05-03
Change boolean validation fields to 'X' enum

mjtravers:
[app#133](https://github.com/fecgov/fecfile-web-app/issues/133)

-


django-backend/fecfiler/f3x_summaries/models.py
```diff
@@ -22,7 +22,7 @@ class F3XSummary(SoftDeleteModel, CommitteeOwnedModel):
     state_of_election = models.TextField(null=True, blank=True)
     coverage_from_date = models.TextField(null=True, blank=True)
     coverage_through_date = models.TextField(null=True, blank=True)
-    qualified_committee = models.BooleanField(default=False, null=True, blank=True)
+    qualified_committee = models.TextField(null=True, blank=True)
```
toddlees, line 25: There has to be another way right?  I know the spec is what we're given but we are going to write error prone code if we use the wrong types.  Every time we are asked to do any logic with fields like these, we're going to have to juggle the types.  If we forget, our code will behave unexpectedly

toddlees (approved):

django-backend/fecfiler/f3x_summaries/models.py
mjtravers: Does the Django models api have an enum? That could be a good middle ground

PR #114, opened 2022-05-05
retrieve fields_to_validate from query params

toddlees:
using query params instead of headers


mjtravers (approved):
Looks good. Passes CR.

PR #121, opened 2022-05-16
Converted memo_code to boolean and added schema query param

mjtravers:
[app#230](https://github.com/fecgov/fecfile-web-app/issues/230)


django-backend/fecfiler/validation/serializers.py
```diff
@@ -38,6 +38,10 @@ def get_schema_name(self, data):
             :py:function:`fecfile_validate.validate` that will
             match a schema defined in the package
         """
+        request = self.context.get("request", None)
+        if request and request.query_params.get("schema"):
+            self.schema_name = request.query_params.get("schema")
```
toddlees, line 43: this got me wondering if setting the schema name would impact subsequent requests.  basically is a single serializer instance used in drf for each request.  I ran a little test printing `schema_name` before and after we set it.  It does not persist across requests.
image

toddlees: so no change required

toddlees (approved):

PR #122, opened 2022-05-16
updated configs for new app on cloud.gov

dheitzer:
## Summary

- Issue number #issue_number

## How to test

-


manifest-dev.yml
```diff
@@ -1,15 +1,15 @@
 ---
 applications:
-- name: fecfile-web-api
+- name: fecfile-web-api-login
```
toddlees, line 3: we won't want to merge this into dev correct?

tasks.py
```diff
@@ -8,7 +8,7 @@
 
 env = cfenv.AppEnv()
 
-APP_NAME = "fecfile-web-api"
+APP_NAME = "fecfile-web-api-login"
```
toddlees, line 11: similar to above. we'll need to change this before it goes into dev?

toddlees (commented):
It's really impressive how much we were able to modularize into the oidc package!

mjtravers (approved):

PR #147, opened 2022-06-17
Boolean migrate hotfix

toddlees:
Improve f3x and schedule a migrations to convert existing "X" and "" values to booleans, because built in conversion had no way to know how to handle those string values.


Elaine-Krauss-TCG (approved):
Looks good

PR #148, opened 2022-06-23
Feature/201 allow users to view existing users

Elaine-Krauss-TCG:
## Summary

 - The test-users fixture is now test-users-extra.json
   - It is also in the committee_account/fixtures directory now
 - A new migration has been created that automatically loads the new fixture

## How to test

After logging in, navigating to the Manage Users page should show multiple users


mjtravers (dismissed):
A couple of change requests:
- Move the test user fixtures and migration of loading the new authentication.account records from the committee account app to the authentication app
- See if Django will accept a fixture of user records that does not have the primary keys declared. Would make the threat of pk collisions not an issue if we can do so.

PR #157, opened 2022-07-01
Feature/262 overlapping coverage dates

dheitzer:
## Summary

- Issue number #issue_number

## How to test

-


django-backend/fecfiler/f3x_summaries/views.py
```diff
@@ -25,6 +27,15 @@ class F3XSummaryViewSet(CommitteeOwnedViewSet):
     ordering_fields = ["form_type", "report_code__label", "coverage_through_date"]
     ordering = ["form_type"]
 
+    @action(detail=False)
+    def coverage_dates(self, request):
+        data = list(
+            self.get_queryset()
```
toddlees, line 33: @dheitzer Because there are filters in the class's queryset (like limiting reports to one's owned by the committtee) I switched this view to use it instead of defining a new queryset.

toddlees (approved):

mjtravers (approved):

PR #162, opened 2022-07-05
Feature/transaction viewset unit tests

Elaine-Krauss-TCG:
## Summary

Adds unit tests to cover the SchATransaction ViewSet.
Also strips primary keys from the models of a fixture that did not need them.


toddlees (approved):
This is going to be SO useful for future tests!

PR #169, opened 2022-07-12
Hotfix - User Management Sorting

Elaine-Krauss-TCG:
The endpoint that provides the list of users for the Manage Users page now sorts on names as "Last, First" instead of "First Last"

The fixture providing dummy users for testing this functionality now has unique last names for every created user.


mjtravers (approved):
Looks good. I was able to manually load the new test data and sort it locally on the front end.

PR #177, opened 2022-07-19
Feature/334 report email confirmation form

Elaine-Krauss-TCG:


django-backend/fecfiler/f3x_summaries/migrations/0009_auto_20220515_2116.py
```diff
@@ -1,10 +1,11 @@
 # Generated by Django 3.2.12 on 2022-05-16 01:16
 
-from fecfiler.f3x_summaries.models import F3XSummary
 from django.db import migrations, models
 
 
 def convert_fecfile_booleans(apps, schema_editor):
+    F3XSummary = apps.get_model("f3x_summaries", "F3XSummary")  # noqa
```
mjtravers, line 7: The caps in this F3XSummary variable is triggering a code smell in SonarCloud. Is there a convincing argument to keep the code smell and not change the variable name?

mjtravers (changes requested):
Just a question about the SonarCloud code smell. Otherwise, PR passes CR.

django-backend/fecfiler/f3x_summaries/migrations/0009_auto_20220515_2116.py
Elaine-Krauss-TCG: Updated accordingly.  Coming up with a variable name that communicates what it's doing was a little tricky, but I think I got it.

Elaine-Krauss-TCG:
> Just a question about the SonarCloud code smell. Otherwise, PR passes CR.

The smell should be taken care of now, thanks!

toddlees (approved):

Issue #124, opened 2022-05-18
Auto-generate transaction IDs when creating transactions


mjtravers:
### Business Reason ### 

As a user creating a new transaction, I want a unique transaction id created when saving.

### Acceptance Criteria ### 

**If** a user saves a transaction
**When** it is saved
**Then** it will be assigned a unique transaction id. 

**And when** I am on the View All Transactions screen
**Then** I can see that the child is associated to a parent transaction using the following columns:
1. Transaction ID (**update to show actual Transaction ID**)
2. Associated with _(**update to show actual Transaction ID of parent**)_

### Tech Notes ###
The Transaction ID will be A/N-20 character limit.
The Transaction ID _must_ be unique within a committee account.
We will generate a UUID then truncate it to 20 characters, and before saving and storing we will ensure that it is not a duplicate of an existing transaction ID. If it is, another shall be generated and the same process followed.
The "repeat" process will max out and throw an error after 5 tries in the unlikely chance it finds duplicate transaction ids 5 times in a row.

### Dev Update ###
On further review with Matt, the UUID will be generated by the API using Python's [UUID.int](https://docs.python.org/3/library/uuid.html#:~:text=The%20UUID%20as%20a%2032%2Dcharacter%20lowercase%20hexadecimal%20string.&text=The%20UUID%20as%20a%20128%2Dbit%20integer.).  This guarantees uniqueness across the database while also being able to be represented by an 8-character hexadecimal value.

- See rules on COL SEQ 4 in the [OFFSET_TO_OPEX spec](https://fecgov.github.io/fecfile-validate/OFFSET_TO_OPEX_spec.html)
- New ID needs to be generated before validation takes place.
- This code can be removed from the front end when this feature is running in the back end: https://github.com/fecgov/fecfile-web-app/blob/feature/230-group-B-transactions/front-end/src/app/transactions/transaction-group-b/transaction-group-b.component.ts#L121-L124

### QA Notes ###
To test, every time a new transaction is created it will be assigned a transaction ID that must be unique across all transactions within the Committee Account. To verify the transaction ID, it will display in the list of transactions table for the report. This transaction ID will not match any other transaction ID for the Committee Account and it will stay the same even if the transaction is edited and updated.

MitchellTCG:
@AureliaKhorsand  
Waiting on more info from 5/19 requirements meeting

AureliaKhorsand:
**Refinement Meeting 6/1/22:** Shelly and Todd not present.

[...]

PR #190, opened 2022-08-05
Feature/124 transaction UUID field

Elaine-Krauss-TCG:
#124 

The `transaction_id` field is now auto-populated with uuid-based 20-char strings of "0-9A-F" values.  This field is indexed and non-editable.
The `back_reference_tran_id_number` field is now the parent transaction's `transaction_id` value.
The `back_reference_sched_name` field is now the parent transaction's `form_type` value.

Paired with [this PR](https://github.com/fecgov/fecfile-web-app/pull/433)


Elaine-Krauss-TCG:
E2E tests caught a critical error related to the serializer updates.  Gimme a few

django-backend/fecfiler/scha_transactions/models.py
```diff
@@ -69,3 +79,39 @@ class SchATransaction(SoftDeleteModel, CommitteeOwnedModel):
...
+    def generate_unique_transaction_id(self):
+        u = uuid.uuid4()
+        hex_id = u.hex.upper()
+        # Take 20 characters from the end, skipping over the 20th char from the right,
+        # which is the version number (uuid4 -> "4")
+        unique_id = hex_id[-21] + hex_id[-19:]
+
+        attempts = 0
+        while SchATransaction.check_for_uid_conflicts(unique_id):
+            unique_id = str(u.hex).upper()[-20:]
```
toddlees, line 100: I think `u.hex` returns the same value every time you call it.  We might need to make a new `uuid.uuid4()` to get a new id

django-backend/fecfiler/scha_transactions/models.py
```diff
@@ -69,3 +79,39 @@ class SchATransaction(SoftDeleteModel, CommitteeOwnedModel):
...
+            attempts += 1
+            if (attempts > 5):
+                logger.info("Unique ID generation failed: Over 5 conflicts in a row")
+                return
+        self.transaction_id = unique_id
+
+    def update_parent_related_values(self, parent):
+        self.back_reference_tran_id_number = parent.transaction_id
```
toddlees, line 108: it might be nice to, at some point take these out of the model and populate them in the queryset.  like `back_reference_tran_id_number=parent_transaction__transaction_id`

django-backend/fecfiler/scha_transactions/test_models.py
```diff
@@ -35,6 +35,31 @@ def test_get_scha_transaction_in_f3x_summary(self):
...
+        self.assertFalse(trans.transaction_id)
+        trans.generate_unique_transaction_id()
+        self.assertTrue(trans.transaction_id)
+
+    def test_catches_uid_conflict(self):
+        trans = SchATransaction.objects.all()[0]
+        existing_uid = trans.transaction_id
+        self.assertTrue(SchATransaction.check_for_uid_conflicts(existing_uid))
+
     def test_save_and_delete(self):
```
toddlees, line 63: can you add an assertion that the `transaction_id` is defined after save?

toddlees (changes requested):

django-backend/fecfiler/scha_transactions/models.py
Elaine-Krauss-TCG: Good catch.  I'm disappointed that I caught that in the migration and didn't also catch it in the model.  Took the opportunity to reduce the code duplication as well by wrapping the whole unique_id generation in its own method.

django-backend/fecfiler/scha_transactions/models.py
Elaine-Krauss-TCG: That would definitely be nice to do at some point

django-backend/fecfiler/scha_transactions/test_models.py
Elaine-Krauss-TCG: That's well worth adding.  Done.

django-backend/fecfiler/scha_transactions/test_models.py
```diff
@@ -49,11 +49,17 @@ def test_generate_uid(self):
         )
         trans = SchATransaction(
             committee_account=committee,
-            report_id=f3x
+            report=f3x
         )
         self.assertFalse(trans.transaction_id)
         trans.generate_unique_transaction_id()
         self.assertTrue(trans.transaction_id)
+        committee.save()
```
toddlees, line 57: I think you had the right idea before not testing save() here.  I meant that you should check that the `transaction_id` is assigned on a saved transaction in the `test_save_and_delete` so we can verify `generate_unique_transaction_id` is being utilized correctly in `save()`

django-backend/fecfiler/scha_transactions/test_models.py
Elaine-Krauss-TCG: Well, once I swapped it from testing `id` to testing `transaction_id`...

django-backend/fecfiler/scha_transactions/test_models.py
toddlees: like at line 72, add `self.assertIsNotNone(transaction_from_db.transaction_id)` this will test that a new transaction will be assigned a transaction_id

django-backend/fecfiler/scha_transactions/test_models.py
```diff
@@ -58,7 +58,10 @@ def test_generate_uid(self):
         f3x.save()
         trans.save()
         saved_trans = SchATransaction.objects.get(id=trans.id)
-        self.assertEquals(trans.id, saved_trans.id)
+        self.assertEquals(trans.transaction_id, saved_trans.transaction_id)
```
toddlees, line 61: This is not quite testing what we want, because you explicitly call `generate_unique_transaction_id` earlier.  we want to test that it's called on `save`

django-backend/fecfiler/scha_transactions/test_models.py
Elaine-Krauss-TCG: The fix is on different lines than the preview that Github is showing, so I'll leave it unresolved till you've reviewed it

django-backend/fecfiler/scha_transactions/test_models.py
Elaine-Krauss-TCG: A check has been added here that expects the saved transaction to have a transaction_id with a length of 20 characters.

django-backend/fecfiler/scha_transactions/test_models.py
Elaine-Krauss-TCG: I think that this is no longer an issue?

toddlees: I was mainly trying to recommend what you implemented on line 72 (so that's all good).

toddlees (approved):

Issue #188, opened 2022-08-01
Update Report Type label for "JULY 31 (MY)" on Mange Reports


AureliaKhorsand:
It was discovered that the STAGE server database had the wrong label for the "MY" report type. A new migration needs to be created to ensure the label in the report_code_labels table for the report_code = 'MY' is set to 'JULY 31 (MY)' so that update will be deployed on DEV.

## Acceptance Criteria ##
When viewing the Report Type for a "MY" report, I will see the text "JULY 31 (MY)" in the "Type of report" column of the reports table

WiseQA:
QA Review Completed.  Verified label change.

Moved to Stage Ready

PR #191, opened 2022-08-05
Updates MY report code label

Elaine-Krauss-TCG:
#188


toddlees (changes requested):
How are you going to make this work with migrations.  Are we going to add an additional migration to load the same JSON?  Maybe we could add a migration that changes MY's label to "JULY 31 (MY)".  It'll be idempotent if the migrations run on a new db and will update a db that ran with the old json

Elaine-Krauss-TCG:
> How are you going to make this work with migrations. Are we going to add an additional migration to load the same JSON? Maybe we could add a migration that changes MY's label to "JULY 31 (MY)". It'll be idempotent if the migrations run on a new db and will update a db that ran with the old json

This has actually been done before [here](https://github.com/fecgov/fecfile-web-api/commit/9d3328249df8238891af36c86e7d44bc0409d403).  I'm pretty sure that when Django runs the existing migration and loads in the fixture, it updates the existing objects without issue.

toddlees:
> > How are you going to make this work with migrations. Are we going to add an additional migration to load the same JSON? Maybe we could add a migration that changes MY's label to "JULY 31 (MY)". It'll be idempotent if the migrations run on a new db and will update a db that ran with the old json
> 
> This has actually been done before [here](https://github.com/fecgov/fecfile-web-api/commit/9d3328249df8238891af36c86e7d44bc0409d403). I'm pretty sure that when Django runs the existing migration and loads in the fixture, it updates the existing objects without issue.

I just tested this locally.  If the migration that loads the fixture has been run, it doesn't get run again.

Elaine-Krauss-TCG:
Migration added that patches the MY Report Code Label instance with the correct value.

toddlees (approved):

PR #198, opened 2022-08-14
Add existence check for report_id in request in ReportViewMixin

mjtravers:
Bug fix for error when editing a transaction.

### DIFF
--- a/django-backend/fecfiler/f3x_summaries/views.py
+++ b/django-backend/fecfiler/f3x_summaries/views.py
@@ -48,7 +48,7 @@ class ReportViewMixin(GenericViewSet):
         report_id = (
             (
                 self.request.query_params.get("report_id")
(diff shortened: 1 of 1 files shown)


django-backend/fecfiler/f3x_summaries/views.py
```diff
@@ -48,7 +48,7 @@ def get_queryset(self):
         report_id = (
             (
                 self.request.query_params.get("report_id")
-                or self.request.data["report_id"]
+                or ("report_id" in self.request.data and self.request.data["report_id"])
```
toddlees, line 51: you could use the same dictionary get `self.request.data.get("report_id"])` as the line above. `("report_id" in self.request.data and self.request.data["report_id"])` will return `report_id` if it's present, but `False` if it's not, which isn't desired

toddlees (changes requested):

PR #218, opened 2022-09-03
Feature/471 app contact lookup prototype

dheitzer:
## Summary

## How to test

-


django-backend/fecfiler/contacts/views.py
```diff
@@ -20,3 +26,45 @@ class ContactViewSet(CommitteeOwnedViewSet):
     in CommitteeOwnedViewSet's implementation of get_queryset()
     """
     queryset = Contact.objects.all().order_by("-id")
+
+    @action(detail=False)
+    def committee_lookup(self, request):
+        q = request.GET.get('q', '').upper()
```
Elaine-Krauss-TCG, line 32: It's a little hard to tell what's going on here

Elaine-Krauss-TCG (approved):
Looks pretty good!  I might add a comment or change `q`'s name to be a little clearer, but I don't think it's at all egregious.

dheitzer:
> Looks pretty good! I might add a comment or change `q`'s name to be a little clearer, but I don't think it's at all egregious.

@Elaine-Krauss-TCG yes I agree it's not the clearest.  I used q for the var name because that's the name of the param from the fec api and I was on the fence whether to name it something different:

https://api.open.fec.gov/v1/names/candidates/?**q=**bid&api_key=mCf2MXteY3H3GcpZmqL1uVTacWEe5eiW3OWFQgBQ

Issue #203, opened 2022-08-17
Fix leak in Git Secrets scan rules


mjtravers:
Update regex patterns for Git Secrets tool to cover leak show below. [Custom Git Secrets rules](https://github.com/fecgov/fecfile-web-api/blob/main/install-git-secrets-hook.sh)

[Click here to see code that slipped past the Git Secrets screener](https://github.com/fecgov/fecfile-web-api/blob/06f1b2bce3c08763bd940d315e74b9b5dbdbe686/django-backend/fecfiler/web_services/tasks.py#L46-L47)

**QA NOTES**
The QA for this ticket is screenshot verification. See the screenshot provided by Elaine in the ticket comment dated Sept. 8 at 11:48am

Elaine-Krauss-TCG:
Updates one of the git secrets patterns to be much more robust:
```['`"]?(dbpasswd|dbuser|dbname|dbhost|api_key|apikey|password|user|guid|hostname|pw|auth)['`"]?\s*[=:]\s*['`".,:\w\/\\+!@$%^&*~?()\[\]{}=-]{4,120}```

Elaine-Krauss-TCG:
Patterns work locally on a test file
![image.png](https://images.zenhubusercontent.com/626ab5e81394ace64f8ed65d/ca484052-c3ae-473f-9d51-b7dbe05bf37f)

WiseQA:
QA Verified by inspection per DEV.

QA Review Completed.  Moved to Stage Ready

PR #226, opened 2022-09-08
[WIP] Updates the custom patterns provided in the git secrets install file

Elaine-Krauss-TCG:
#203 
  
Regex rules updated to include 
```
(dbpasswd|dbuser|dbname|dbhost|api_key|apikey|password|guid|hostname|pw).*[=:][^(?:>|.*=>|.*>$)]
(DBPASSWD|DBUSER|DBNAME|DBHOST|API_KEY|APIKEY|PASSWORD|GUID|HOSTNAME|PW).*[=:][^(?:>|.*=>|.*>$)]
(user|auth|USER|AUTH)\s*[=:][^(?:>|.*=>|.*>$)]
```

It breaks down to:

`(dbpasswd|dbuser|dbname|dbhost|api_key|apikey|password|guid|hostname|pw)`
finds a target phrase

`.*[=:]`
that is to the *left* of an assignment

`[^(?:>|.*=>|.*>$)]`
but that is *not* followed by either
1) immediately a '>' (such as =>)
2) a later '=>'
or
3) the line ends with '>' (such as in HTML)

'user' and 'auth' have stricter checks matching only whitespace between them and the assignment

### DIFF
--- a/install-git-secrets-hook.sh
+++ b/install-git-secrets-hook.sh
@@ -40,8 +40,9 @@ fi
 git secrets --register-aws $GLOBAL_TOKEN
 
 # Add general custom rules
-git secrets --add $GLOBAL_TOKEN '(dbpasswd|dbuser|dbname|dbhost|api_key|apikey|password|user|guid|hostname|pw|auth)\s*[=:]\s*['"'"'0-9a-zA-Z_\/+!{}=-]{4,120}'
-git secrets --add $GLOBAL_TOKEN '(DBPASSWD|DBUSER|DBNAME|DBHOST|API_KEY|APIKEY|PASSWORD|USER|GUID|HOSTNAME|PW|AUTH)\s*[=:]\s*['"'"'0-9a-zA-Z_\/+!{}=-]{4,120}'
+git secrets --add $GLOBAL_TOKEN '(dbpasswd|dbuser|dbname|dbhost|api_key|apikey|password|guid|hostname|pw).*[=:][^(?>|.*=>|.*>$)]'
+git secrets --add $GLOBAL_TOKEN '(DBPASSWD|DBUSER|DBNAME|DBHOST|API_KEY|APIKEY|PASSWORD|GUID|HOSTNAME|PW).*[=:][^(?>|.*=>|.*>$)]'
+git secrets --add $GLOBAL_TOKEN '(user|auth|USER|AUTH)\s*[=:][^(?>|.*=>|.*>$)]'
 git secrets --add $GLOBAL_TOKEN '(aws_access_key_id|aws_secret_access_key)\s*[=:]\s*['"'"'0-9a-zA-Z\/+]{20,42}'
 git secrets --add $GLOBAL_TOKEN '(AWS_ACCESS_KEY_ID|AWS_SECRET_ACCESS_KEY)\s*[=:]\s*['"'"'0-9a-zA-Z\/+]{20,42}'
 


toddlees (approved):

install-git-secrets-hook.sh
```diff
@@ -40,8 +40,9 @@ fi
 git secrets --register-aws $GLOBAL_TOKEN
 
 # Add general custom rules
-git secrets --add $GLOBAL_TOKEN '(dbpasswd|dbuser|dbname|dbhost|api_key|apikey|password|user|guid|hostname|pw|auth)\s*[=:]\s*['"'"'0-9a-zA-Z_\/+!{}=-]{4,120}'
-git secrets --add $GLOBAL_TOKEN '(DBPASSWD|DBUSER|DBNAME|DBHOST|API_KEY|APIKEY|PASSWORD|USER|GUID|HOSTNAME|PW|AUTH)\s*[=:]\s*['"'"'0-9a-zA-Z_\/+!{}=-]{4,120}'
```
toddlees, line 44: why do we want to exclude html?

Elaine-Krauss-TCG: There were a lot of false positives in the html code, though it was mainly with "user" and "auth."  My thought was that the only thing we're really concerned with is assignment; we don't want someone submitting password="TheActualPassword" or something along that line.  I can't think of any context in which that would be done in html code.

toddlees (changes requested):
`.*` between the left side value, assignment operator, and right hand value might be too broad.

image

maybe we could limit it to `["'`\s]*`

the `.*` on the right hand side is matching everything that follows, meaning the regex code after it is effectively only matching the last character.  When you change `.*` (to the right of the assignment operator) to something stricter, it only matches one character after the operator

image

to fix this, you can re-add the {4,120} to match the right hand value that's longer than 4 chars and not longer than 120
image

Elaine-Krauss-TCG:
`.*` might be too broad

The first iteration I had of the regex changes actually did what you're suggesting, I even had a version of `["'\s]*`, but this only catches variables that _end_ with user, password, etc.  It is plausible that a variable name would contain some additional text between the triggering phrase and the assignment.  IMO, it's better to have false positives than to let something slip past.

As far as the right side goes,

the only thing that I think we should care about is whether or not an assignment is taking place.  _What_ is being assigned is irrelevant, and matching on it is an opportunity for something to slip past.  At a minimum, we would need to expand the character matching to include every symbol that could ever possibly be used in a password or a secret key, such as the '@' that was in the password that slipped through [before](https://github.com/fecgov/fecfile-web-api/blob/06f1b2bce3c08763bd940d315e74b9b5dbdbe686/django-backend/fecfiler/web_services/tasks.py#L46-L47)

toddlees (approved):

toddlees:
> As far as what's left of the assignment goes,
> 
> The first iteration I had of the regex changes actually did what you're suggesting, I even had a version of `["'\s]*`, but this only catches variables that _end_ with user, password, etc. It is plausible that a variable name would contain some additional text between the triggering phrase and the assignment. IMO, it's better to have false positives than to let something slip past.
> 
> As far as the right side goes,
> 
> the only thing that I think we should care about is whether or not an assignment is taking place. _What_ is being assigned is irrelevant, and matching on it is an opportunity for something to slip past. At a minimum, we would need to expand the character matching to include every symbol that could ever possibly be used in a password or a secret key, such as the '@' that was in the password that slipped through [before](https://github.com/fecgov/fecfile-web-api/blob/06f1b2bce3c08763bd940d315e74b9b5dbdbe686/django-backend/fecfiler/web_services/tasks.py#L46-L47)

You've convinced me 👍

mjtravers (approved):
This looks good. I was able to re-install the git secrets configuration locally and it caught the use case in the ticket when running a test commit.

The dependency failure will be corrected when merged into the develop branch as the exception file has been updated there

PR #227, opened 2022-09-13
Feature/72 individual lookup

dheitzer:
[app#72](https://github.com/fecgov/fecfile-web-app/issues/72)

### DIFF
--- a/django-backend/fecfiler/contacts/views.py
+++ b/django-backend/fecfiler/contacts/views.py
@@ -12,6 +12,10 @@ import logging
 
 logger = logging.getLogger(__name__)
 
+default_max_fec_results = 10
+default_max_fecfile_results = 10
+max_allowed_results = 100
+
 
 class ContactViewSet(CommitteeOwnedViewSet):
     """
@@ -34,18 +38,12 @@ class ContactViewSet(CommitteeOwnedViewSet):
         if q is None:
             return HttpResponseBadRequest()
 
-        max_fec_results = 10
-        max_fec_results_param = request.GET.get("max_fec_results", "")
-        if max_fec_results_param is not None and max_fec_results_param.isnumeric():
-            max_fec_results = int(max_fec_results_param)
+        max_fec_results = self.get_int_param_value(
+            request, "max_fec_results", default_max_fec_results, max_allowed_results)
 
-        max_fecfile_results = 10
-        max_fecfile_results_param = request.GET.get("max_fecfile_results", "")
-        if (
-            max_fecfile_results_param is not None
-            and max_fecfile_results_param.isnumeric()
-        ):
-            max_fecfile_results = int(max_fecfile_results_param)
+        max_fecfile_results = self.get_int_param_value(
+            request, "max_fecfile_results", default_max_fecfile_results,
+            max_allowed_results)
 
         query_params = urlencode(
             {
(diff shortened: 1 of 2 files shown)


django-backend/fecfiler/contacts/views.py
```diff
@@ -61,14 +62,39 @@ def committee_lookup(self, request):
...
+        max_fecfile_results_param = request.GET.get("max_fecfile_results", "")
+        if (
+            max_fecfile_results_param is not None
+            and max_fecfile_results_param.isnumeric()
+        ):
+            max_fecfile_results = int(max_fecfile_results_param)
+
+        fecfile_individuals = list(
+            self.get_queryset()
+            .filter(Q(type__icontains="IND") & (
```
mjtravers, line 91: I believe this is an exact match with the field value so doesn't need a "contains". Would this work and be more exact?: .filter(Q(type="IND") & ...

django-backend/fecfiler/contacts/views.py
```diff
@@ -61,14 +62,39 @@ def committee_lookup(self, request):
...
 
         return JsonResponse(return_value)
+
+    @action(detail=False)
+    def individual_lookup(self, request):
+        q = request.GET.get("q")
+        if q is None:
+            return HttpResponseBadRequest()
+
+        max_fecfile_results = 10
```
mjtravers, line 81: Let's make this a constant outside the class and use that constant in both individual_lookup() and committee_lookup() functions

mjtravers (changes requested):
A couple of comments put inline.

Elaine-Krauss-TCG (approved):

mjtravers (approved):

toddlees (approved):

dheitzer:
Updates are ready for  CR.

django-backend/fecfiler/contacts/views.py
```diff
@@ -76,14 +79,23 @@ def individual_lookup(self, request):
         if q is None:
             return HttpResponseBadRequest()
 
+        tokens = list(filter(None, re.split('[^\\w+]', q)))
+        if (not tokens) or (len(tokens) < 1):
+            return HttpResponseBadRequest()
+        term = ('.*' + '.* .*'.join(tokens) + '.*').lower()
```
toddlees, line 83: I believe this brings us back into the world of the upper/lower problem.  I personally don't want to be in the business of designing search engines and I'd recommend that we stick to already developed solutions.

toddlees (changes requested):

django-backend/fecfiler/contacts/views.py
toddlees: I believe you mentioned before @dheitzer that adding complexity to the querystring parsing proportionally adds unexpected behavior

Elaine-Krauss-TCG (approved):

django-backend/fecfiler/contacts/views.py
dheitzer: Yes that's a good point @toddlees.  I thought I would need to change to this based on discussion with @mjtravers and seeing how fec.gov seems to handle delimiters and Aurelia's feedback but I'm open to however we want to handle this.  Maybe we can discuss/review Tues when Matt's back?

toddlees: image
image

Splitting up the search by tokens seems to make names with apostrophe's unmatchable.

toddlees: It looks like fec splits up the tokens like you've got, so it's probably good that we match it in that sense.  If we can implement that without trading off matching names with `'`, that might be the way forward.

toddlees: In a meeting with @MitchellTCG  we decided that we will pass on matching names with `'` until later

toddlees (approved):

Issue #117, opened 2022-05-08
Convert primary key to UUID for f3x_summaries and scha_transactions


mjtravers:
In the future, the database will have multiple tables capturing different Form types and multiple tables capturing different Schedule types. The primary key ids must be unique across all the tables so we can readily retrieve a form or transaction by ID only.

**Acceptance Criteria**
- Table f3x_summaries has a UUID for primary key
- Table scha_transactions has a UUID for primary key
- Site passes all unit tests and regression tests

WiseQA:
QA Review as a Visual inspection with DEV verifying Unit Tests.

QA Review Completed.  Moved to Stage Ready.

PR #229, opened 2022-09-15
Feature/117 migrate to UUID

toddlees:
#117


django-backend/fecfiler/committee_accounts/migrations/0002_committeeaccount_uuid.py
```diff
@@ -0,0 +1,32 @@
...
+        ("committee_accounts", "0001_initial"),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name="committeeaccount",
+            name="uuid",
+            field=models.UUIDField(default=uuid.uuid4, editable=False),
+        ),
+        migrations.RunPython(create_uuid),
```
Elaine-Krauss-TCG, line 26: You should probably define a reverse function here even if it's just a `noop`

Elaine-Krauss-TCG (approved):
I was able to both run the migrations on an existing db and to spin up a fresh db without any issues.  The code looks great, though I might consider defining reverse functions on the RunPython instructions, even if they just wind up being [noop](https://stackoverflow.com/questions/66479161/what-does-runpython-noop-do)'s

dheitzer (approved):

PR #233, opened 2022-09-27
460 earmark save

toddlees:
Can query for transaction based on parent id

PR #250, opened 2022-10-18
460 Add child transactions to api GET output

mjtravers:
#233

### DIFF
--- a/django-backend/fecfiler/scha_transactions/serializers.py
+++ b/django-backend/fecfiler/scha_transactions/serializers.py
@@ -85,8 +85,6 @@ class SchATransactionParentSerializer(SchATransactionSerializer):
         required=False,
     )
 
-    parent_transaction = SchATransactionSerializer(required=False, allow_null=True)
-
     def to_representation(self, instance):
         ret = super().to_representation(instance)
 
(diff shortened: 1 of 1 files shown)


django-backend/fecfiler/scha_transactions/serializers.py
```diff
@@ -80,12 +80,25 @@ class Meta:
...
+
+    def to_representation(self, instance):
+        ret = super().to_representation(instance)
+
+        # Add child transactions to transaction.children field in JSON output
+        existing_children = SchATransaction.objects.filter(
+            parent_transaction_id=instance.id
+        ).all()
+        ret["children"] = map(
+            lambda child: SchATransactionParentSerializer(child).data, existing_children
```
toddlees, line 98: I think you can use self.to_representation(child) here for an exact match of functionality
`map(self.to_representation, existing_children)`

toddlees (approved):

django-backend/fecfiler/scha_transactions/serializers.py
mjtravers: Good tip. I made the change and it works.

Elaine-Krauss-TCG (approved):

PR #259, opened 2022-10-28
Feature/119 dot fec add report level memo

dheitzer:
## Summary

## How to test

-


toddlees (approved):
Looks great!  You could add a test to test_dot_fec_serializer to ensure the row looks right

dheitzer:
> Looks great! You could add a test to test_dot_fec_serializer to ensure the row looks right

nice catch, added a test there as well.

PR #262, opened 2022-11-01
Feature/403 memo text as separate record

Elaine-Krauss-TCG:
App#[403](https://app.zenhub.com/workspaces/fecfile-online-619e578e68408b001c831251/issues/fecgov/fecfile-web-app/403)


django-backend/fecfiler/scha_transactions/migrations/0033_memo_description_to_memo_record.py
```diff
@@ -0,0 +1,38 @@
...
+from django.db import migrations
+
+
+def text_field_to_text_record(apps, _):
+    sch_a_transaction = apps.get_model("scha_transactions", "schatransaction")  # noqa
+    memo_text = apps.get_model("memo_text", "memotext")
+    for transaction in sch_a_transaction.objects.all():
+        transaction.memo_text = memo_text()
+        transaction.memo_text.text4000 = transaction.memo_text_description
+        transaction.memo_text.transaction_id_number = transaction.id
```
toddlees, line 12: In this case the `transaction_id_number`... you're going to love this... is the "transaction" id of the memo, as in a unique id for the memo record, whereas the `back_reference_tran_id_number` is the (FEC, not internal fecfile) id of the transaction it's tied to.  so you want to assign `transaction.memo_text.back_reference_tran_id_number = transaction.transaction_id` and `transaction.memo_text.transaction_id_number = <need to look up convention for this>`

toddlees (changes requested):

django-backend/fecfiler/scha_transactions/migrations/0033_memo_description_to_memo_record.py
toddlees: Plus  `transaction.memo_text.back_reference_sched_form_name = transaction.form_type`

toddlees (changes requested):

django-backend/fecfiler/scha_transactions/models.py
```diff
@@ -74,7 +74,7 @@ class SchATransaction(SoftDeleteModel, CommitteeOwnedModel, ReportMixin):
     conduit_state = models.TextField(null=True, blank=True)
     conduit_zip = models.TextField(null=True, blank=True)
     memo_code = models.BooleanField(null=True, blank=True, default=False)
-    memo_text_description = models.TextField(null=True, blank=True)
```
toddlees, line 77: This will impact how the transaction will be serialized in the .fec.  (a design that is showing it's flaws by being brittle when we want to change the model).  You should leave this field as is and just don't populate it

django-backend/fecfiler/scha_transactions/models.py
```diff
@@ -133,5 +137,8 @@ def save(self, *args, **kwargs):
             self.transaction_id = self.generate_unique_transaction_id()
         if not self.back_reference_tran_id_number and self.parent_transaction:
             self.update_parent_related_values(self.parent_transaction)
+        if self.memo_text:
+            self.memo_text.transaction_id = self.id
```
toddlees, line 141: is this necessary if there is already a fk relation between the tables with transaction.memo_text?

django-backend/fecfiler/scha_transactions/migrations/0033_memo_description_to_memo_record.py
Elaine-Krauss-TCG: Do we want to save any of that, or do we want to retrieve it from the related transaction?  I'm pretty against saving things twice wherever possible

django-backend/fecfiler/scha_transactions/serializers.py
```diff
@@ -150,6 +158,20 @@ def create_or_update_contact(self, validated_data: dict):
         else:
             Contact.objects.filter(id=tran_contact_id).update(**contact_data)
 
+    def create_or_update_memo_text(self, validated_data: dict):
+        memo_data = validated_data.pop("memo_text", None)
+        tran_memo_text_id = validated_data.get("memo_text_id", None)
+        if not tran_memo_text_id:
+            if not memo_data:
+                raise ValidationError(
+                    {"memo_text": ["No memo_text or memo_text_id provided"]}
```
toddlees, line 167: Unlike Contacts, I think we want to allow transactions to be saved if they don't have a memo_text attached.  it should be optional in many cases

django-backend/fecfiler/shared/test_transactions.py
```diff
@@ -0,0 +1,25 @@
+from django.test import TestCase
```
toddlees, line 1: I would leave this out for now until we have a clear path forward for dividing up the schedules on the api

django-backend/fecfiler/scha_transactions/models.py
Elaine-Krauss-TCG: Oops, I had removed it and then *put it back*, but I guess I put it back the wrong way...

django-backend/fecfiler/scha_transactions/models.py
Elaine-Krauss-TCG: This is likely going to be helpful in the future when there are many different transaction models and tables.  This field will be universal, pointing to the id of _a_ transaction in one of the tables.  You'll be able to use the new helper function in the shared app to find the transaction instead of checking for which foreign key field is populated.

django-backend/fecfiler/scha_transactions/migrations/0033_memo_description_to_memo_record.py
toddlees: me too.  you could leave the backreference fields till the .fec serialization, but transaction_id_number is information that only pertains to the memo_text record and can't be extrapolated from the transaction it refers to

django-backend/fecfiler/scha_transactions/serializers.py
Elaine-Krauss-TCG: The front-end always sends a value for memo_text, it's just that the text4000 field is sometimes empty.  This is because we don't want to create a new transaction without creating a memo_text record.  I was on the fence about that, but ultimately I don't think we want to be deleting records every time someone edits a memo to empty.

django-backend/fecfiler/scha_transactions/migrations/0033_memo_description_to_memo_record.py
Elaine-Krauss-TCG: Hmm...  That's a really good point.  It would naturally be a value derived from our UUID's for consistency's sake.  I kind of think we should define a method in the transaction model that gets the FEC-suitable id from the UUID?  Maybe?  Idk...

toddlees: I think that's a good idea, all that matters is that it's unique within the report (across amendments).  like the memo's uuid, but truncated

django-backend/fecfiler/scha_transactions/serializers.py
toddlees: I see what you're getting at with clearing out the field.  I think the cost is pretty high for simplifying that case.  I think the edge case where they clear it out can be left alone and optimized to where we delete empty ones in the future.

There are pretty big ramifications to having a memo_text record for every transaction.  One is that the .fec will have memo_text record per transaction, which will be almost on the order of doubling its size (if memo's are rare).  Disclosures (pdf versions) will have a lot of empty pages cluttering the document.  We'll have a whole lot more records in the db than what we need.

django-backend/fecfiler/web_services/dot_fec/test_dot_fec_serializer.py
```diff
@@ -14,6 +14,7 @@ class DotFECSerializerTestCase(TestCase):
         "test_committee_accounts",
```
toddlees, line 14: Not sure where to put this comment. You'll need to update `compose_report_level_memos` in the dot_fec_composer to only return report memos and update the transactions section to insert transaction memos after their transactions

toddlees (changes requested):

django-backend/fecfiler/scha_transactions/migrations/0033_memo_description_to_memo_record.py
Elaine-Krauss-TCG: This is already being done, thankfully.  On generation, each transaction creates a transaction_id, which is a 20 character string derived from a UUID that is checked for uniqueness against all other transactions.  We use that for the FEC suitable id that is shown on the front-end, and there's no reason not to use it here.

toddlees: for transactions yeah, but the `transaction_id_number` is called what it's called because the .fec spec didn't change the name of the field for the memo spec.  If I were to name it, I'd call it something like "memo_id".

toddlees (approved):

django-backend/fecfiler/memo_text/models.py
```diff
@@ -15,10 +19,55 @@ class MemoText(SoftDeleteModel, CommitteeOwnedModel, ReportMixin):
...
+            report_id=self.report_id, transaction_uuid=None
+        )
+        return "REPORT_MEMO_TEXT" + str(len(report_memos) + 1)
+
+    def generate_uid(self):
+        unique_id = uuid.uuid4()
+        hex_id = unique_id.hex.upper()
+        # Take 20 characters from the end, skipping over the 20th char from the right,
+        # which is the version number (uuid4 -> "4")
+        return hex_id[-21] + hex_id[-19:]
```
mjtravers, line 51: This method, generate_uid(),  is the exact same as the one of same name in the scha_transactions model. Let's put it in the new "shared" app that contains the common methods across apps/models.

django-backend/fecfiler/shared/transactions.py
```diff
@@ -0,0 +1,14 @@
...
+    tables = [SchATransaction]
+    matches = []
+    for table in tables:
+        matches += list(table.objects.filter(id=uuid))
+
+    if len(matches) > 1:
+        raise LookupError
+    if len(matches) == 1:
+        return matches[0]
+    return None
```
mjtravers, line 14: This is great!

mjtravers (changes requested):
One small quibble about eliminating duplicate code.

@Elaine-Krauss-TCG @toddlees @dheitzer 
Not for this PR, but it'd be good to discuss alternatives to where we keep the create_or_update_memo_text() and create_or_update_contact() methods because we'll need them for other schedule types. Should they go in the new shared app... in a mixin... in the manager for the MemoText and Contact models? Food for thought.

django-backend/fecfiler/memo_text/models.py
Elaine-Krauss-TCG: Great call; taken care of

django-backend/fecfiler/shared/transactions.py
Elaine-Krauss-TCG: :+1:

mjtravers (approved):

PR #298, opened 2022-12-12
Feature/207 django 4.1

mjtravers:


django-backend/fecfiler/authentication/views.py
```diff
@@ -37,39 +51,170 @@ class AccountViewSet(GenericViewSet, ListModelMixin):
...
+                )
+            )
+            .filter(cmtee_id=self.request.user.cmtee_id)
+            .all()
+        )
 
         return queryset
 
 
+# class AccountViewSet(viewsets.ModelViewSet):
```
dheitzer, line 72: May want to remove this

requirements.txt
```diff
@@ -1,30 +1,30 @@
...
-django-cors-headers==3.11.0
-django-storages==1.12.3
-djangorestframework==3.13.1
-djangorestframework-jwt==1.11.0
-drf-spectacular==0.21.2
+Django==4.1.3
+django-cors-headers==3.13.0
+django-storages==1.13.1
+djangorestframework==3.14.0
+djangorestframework-simplejwt[crypto]==5.2.2
```
dheitzer, line 13: is simplejwt still needed?

dheitzer:
It looks like test_authentication.py and test_token.py have some unresolved imports that we might want to take care of as part of this change.

dheitzer:
> It looks like test_authentication.py and test_token.py have some unresolved imports that we might want to take care of as part of this change.

fixed

django-backend/fecfiler/authentication/views.py
```diff
@@ -37,39 +51,136 @@ class AccountViewSet(GenericViewSet, ListModelMixin):
...
-                               domain=FFAPI_COOKIE_DOMAIN)
+        response.delete_cookie(
+            FFAPI_COMMITTEE_ID_COOKIE_NAME, domain=FFAPI_COOKIE_DOMAIN
+        )
+        response.delete_cookie(FFAPI_EMAIL_COOKIE_NAME, domain=FFAPI_COOKIE_DOMAIN)
+        response.delete_cookie("csrftoken", domain=FFAPI_COOKIE_DOMAIN)
         return response
+
+
+class IsAccountOwner(permissions.BasePermission):
```
toddlees, line 102: Is this used anywhere?

toddlees (changes requested):

django-backend/fecfiler/authentication/views.py
```diff
@@ -37,39 +51,136 @@ class AccountViewSet(GenericViewSet, ListModelMixin):
...
+                )
+            )
+            .filter(cmtee_id=self.request.user.cmtee_id)
+            .all()
+        )
 
         return queryset
 
 
 class LoginDotGovSuccessSpaRedirect(View):
```
toddlees, line 72: What do we get from having this in a separate class, rather than making it an action in the AccountViewSet.  Same with `LoginDotGovSuccessLogoutSpaRedirect`, `LogoutView`, and `authenticate_login` (a separate api_view rather than a class)

toddlees (changes requested):

django-backend/fecfiler/authentication/views.py
toddlees: https://www.django-rest-framework.org/api-guide/viewsets/#viewset

django-backend/fecfiler/authentication/views.py
mjtravers: This class is not used. It has been removed from the code.

django-backend/fecfiler/authentication/views.py
mjtravers: Will make this refactor in a later PR.

Elaine-Krauss-TCG (approved):

dheitzer (approved):

toddlees (approved):

PR #312, opened 2022-12-19
Login dot gov fix for logout endpoint returning 403

dheitzer:
## Summary

## How to test

-

### DIFF
--- a/django-backend/fecfiler/authentication/views.py
+++ b/django-backend/fecfiler/authentication/views.py
@@ -118,6 +118,13 @@ def handle_invalid_login(username):
     )
 
 
+def delete_user_logged_in_cookies(response):
+    response.delete_cookie(FFAPI_COMMITTEE_ID_COOKIE_NAME, domain=FFAPI_COOKIE_DOMAIN)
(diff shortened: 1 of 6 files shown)


toddlees (approved):

django-backend/fecfiler/authentication/views.py
```diff
@@ -140,11 +147,10 @@ def login_redirect(request):
 
 @api_view(["GET"])
 @require_http_methods(["GET"])
+@permission_classes([])
```
mjtravers, line 150: I'm wondering if we need @authentication_classes([]) here as well. When testing authenticate_login() and authenticate_logout() functions, they were getting authentication denied responses without that decorator.

dheitzer: @mjtravers yes I wondered about this too, however I think since we are no longer using 
`rest_framework_jwt.authentication.JSONWebTokenAuthentication` in our `DEFAULT_AUTHENTICATION_CLASSES` that this is no longer needed.  I tested without it and it seems to work, so I opted to leave it out.

mjtravers: Sounds good.

mjtravers (approved):

PR #333, opened 2023-01-31
feature/206 .fec header issue

dheitzer:
We were getting the following error in our .fec headers which is causing webcheck to fail:

HDR^\FEC^\8.4^\FECFile Online^\0.0.1^\^\\**(None,)**^\

-

### DIFF
--- a/django-backend/fecfiler/web_services/dot_fec/dot_fec_composer.py
+++ b/django-backend/fecfiler/web_services/dot_fec/dot_fec_composer.py
@@ -80,7 +80,7 @@ class Header:
         self.soft_name = soft_name
         self.soft_ver = soft_ver
         self.rpt_id = rpt_id
-        self.rpt_number = (rpt_number,)
+        self.rpt_number = rpt_number
         self.hdrcomment = hdrcomment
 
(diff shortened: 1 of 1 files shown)


Elaine-Krauss-TCG (approved):

django-backend/fecfiler/web_services/dot_fec/dot_fec_composer.py
```diff
@@ -80,7 +80,7 @@ def __init__(
         self.soft_name = soft_name
         self.soft_ver = soft_ver
         self.rpt_id = rpt_id
-        self.rpt_number = (rpt_number,)
+        self.rpt_number = rpt_number
```
toddlees, line 83: very interesting!  i wonder how that happened

toddlees (approved):

django-backend/fecfiler/web_services/dot_fec/dot_fec_composer.py
dheitzer: same.  my first thought was eslint got it somehow?

PR #334, opened 2023-01-31
Feature/601 Repairs sorting on the contacts management page

Elaine-Krauss-TCG:
[App ticket #601](https://app.zenhub.com/workspaces/fecfile-online-619e578e68408b001c831251/issues/gh/fecgov/fecfile-web-app/601)

### DIFF
--- a/django-backend/fecfiler/contacts/views.py
+++ b/django-backend/fecfiler/contacts/views.py
@@ -37,7 +37,13 @@ class ContactViewSet(CommitteeOwnedViewSet):
     queryset = (
         Contact.objects
         .annotate(transaction_count=Count("transaction"))
-        .annotate(sort_name=Concat("name", "last_name", Value(" "), "first_name", output_field=CharField()))
+        .annotate(sort_name=Concat(
+            "name",
+            "last_name",
+            Value(" "),
+            "first_name",
+            output_field=CharField())
+        )
         .all()
     )
     filter_backends = [filters.OrderingFilter]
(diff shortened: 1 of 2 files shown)


mjtravers (changes requested):
This is not sorting as expected. When I have a mix of contact types, the sort is grouping or not sorting on certain types. Sorting simply by "name" for the first column is not working.

django-backend/fecfiler/contacts/views.py
```diff
@@ -32,11 +33,24 @@ class ContactViewSet(CommitteeOwnedViewSet):
     The queryset will be further limmited by the user's committee
     in CommitteeOwnedViewSet's implementation of get_queryset()
     """
+
     queryset = (
-        Contact.objects.annotate(transaction_count=Count("transaction"))
+        Contact.objects
+        .annotate(transaction_count=Count("transaction"))
+        .annotate(sort_name=Concat("name", "last_name", Value(" "), "first_name", output_field=CharField()))
```
toddlees, line 40: if you use alias for sort_name, you won't need to remove it in the serializer because it will only exist during the query

toddlees (changes requested):

django-backend/fecfiler/contacts/views.py
toddlees, line 40: lint is saying this line is too long

toddlees (changes requested):

django-backend/fecfiler/contacts/views.py
Elaine-Krauss-TCG: Good catch

toddlees (approved):

mjtravers (approved):

PR #342, opened 2023-02-09
Feature/206 test committee accounts for filing

dheitzer:
created 'mock_openfec' app to go between the client and the openfec api and also to allow for mock committee data to support filing.  This service was implemented with the goal of minimal changes to the client.


Elaine-Krauss-TCG (approved):

dheitzer:
Made the changes discussed in our dev meeting to:

1) use an environment variable to determine committee id data to override
2) use a local json file to override the committee data for specified committee ids

PR #390, opened 2023-04-03
Feature/582 update future transactions on contact edit

dheitzer:
[app#582](https://github.com/fecgov/fecfile-web-app/issues/582)


django-backend/fecfiler/transactions/schedule_a/serializers.py
toddlees, line 1: I think the solution to this ticket needs to address any schedule that has a contact.  That might mean that it makes more sense to trigger the propagation in either the LinkedContactSerializerMixin or ContactSerializer

django-backend/fecfiler/transactions/schedule_a/serializers.py
```diff
@@ -144,7 +150,19 @@ def update(self, instance, validated_data: dict):
...
                     setattr(instance.schedule_a, attr, value)
             instance.schedule_a.save()
-            return super().update(instance, validated_data)
+            updated = super().update(instance, validated_data)
+            self.execute_update_future_transaction_contacts_task_if_needed(
+                validated_data.get('contact_id'), schedule_a_data
+            )
+            return updated
+
+    def execute_update_future_transaction_contacts_task_if_needed(
```
toddlees, line 159: What if instead of going to each transaction and juggling whether or not it's been updated then firing off a new task, can we just on contact update, find/query all subsequent transactions and call update on that queryset.

toddlees (changes requested):

django-backend/fecfiler/transactions/schedule_a/serializers.py
toddlees: One way to do this would be to put a function in LinkedContactSerializerMixin called get_effective_date() which returns None.  Then have ScheduleASerializer and ScheduleBSerializer implement effective_date() and return the value of contribution_date and expenditure_date respectively:
```
'''ScheduleBSerializer'''
def get_effective_date(self, validated_data):
  return validated_data.get('expenditure_date')

'''ScheduleASerializer'''
def get_effective_date(self, validated_data):
  return validated_data.get('contribution_date')
```

the LinkedContactSerializerMixin code would look like this:
```
if not contact_id:
  ...
else:
  contacts = Contact.objects.filter(id=contact_id)
  contact = contact_id_query_set.first()
  fields_that_have_changed = [k for k, v in contact_data.items() if getattr(contact,k) != v]
  if fields_that_have_changed:
    contacts.update(**contact_data)
    effective_date = self.get_effective_date()
    if effective_date:
      update_future_transaction_contacts.s(contact_id, effective_date).apply_async(retry=False)
```
Then the task would query `Transactions.objects.filter(date__gte=effective_date)`

toddlees (changes requested):

django-backend/fecfiler/transactions/schedule_a/serializers.py
dheitzer: > I think the solution to this ticket needs to address any schedule that has a contact. That might mean that it makes more sense to trigger the propagation in either the LinkedContactSerializerMixin or ContactSerializer

I have added schedule b.  Per our dev discussion (and confirmation with Aurelia) we will be creating an additional ticket for schedule c due to the transaction date complications with that schedule.

dheitzer: > One way to do this would be to put a function in LinkedContactSerializerMixin called get_effective_date() which returns None. Then have ScheduleASerializer and ScheduleBSerializer implement effective_date() and return the value of contribution_date and expenditure_date respectively:

I oped for a similar approach by having the schedule serializers specify the transaction date field.  I felt this was a clean approach since they also have to specify the contact fields for this functionality but I'm open if we want to change this.

> the LinkedContactSerializerMixin code would look like this:

That makes sense.  I have almost exactly that in the LinkedContactSerializerMixin create_or_update_contact function.

dheitzer:
In this iteration:

- Removed the Celery task as discussed
- Moved update functionality into Transaction serializer generic to be handled by any schedule
- Opted to keep the LinkedContactSerializerMixin check for contact update and propagation to schedule serializers, specifically to handle the case of transaction creation with updated contact.

dheitzer:
note:  as part of this ticket, I updated schedule_b database column names with "beneficiary" to match the Schb spec.

Elaine-Krauss-TCG (approved):

toddlees (approved):

mjtravers (approved):

PR #415, opened 2023-04-25
Feature/449 452

toddlees:
app#449
app#452


Elaine-Krauss-TCG (approved):
I've tested it and it works, but I do want to point out that there's a noticeable performance hit when saving Tier 3 transactions.  I have no suggestions for how this might be improved, though.

toddlees:
> I've tested it and it works, but I do want to point out that there's a noticeable performance hit when saving Tier 3 transactions. I have no suggestions for how this might be improved, though.

I notice the same hit.  This absolutely calls for another look at how we're doing serialization.  3 tiers of transactions should be a drop in the bucket performance-wise, but something is definitely lagging.

dheitzer (approved):

mjtravers (approved):

PR #442, opened 2023-05-23
Feature/917+918 DSP lines 28 and 34

dheitzer:
I created the function `_create_expenditure_sum()` in `summary.py`, but I wonder if we can just rename the existing function `_create_contribution_sum()` to something more generic since "amount" is always present in the transaction model.  

What do you all think?


toddlees (approved):
I think that's a good call to just use one function called `create_amount_sum`. We'd only want different functions if we expect them to ever behave differently and i don't see that happening

Issue #400, opened 2023-04-10
Research: Find a way to create a DCF file


mjtravers:
To facilitate migrating from the the FECfile Desktop to FECfile Online, it is necessary to be able to import a DCF formatted file created by the desktop. There is currently no documentation available on the DCF format so we need to reverse engineer how a DCF file is made so that we can import them into Fecfile Online.

**Acceptance Criteria**
As a user, I would like to use the DCF file I have on FECFile Desktop to migrate FEC reports into FECFile Online.

Elaine-Krauss-TCG:
I've got very good news, and that's that .DCf files are pretty much just CSV files.  Skimming through the examples that Ken Lally provided, I see no sign of any special characters, and the formatting seems to be fairly straightforward.

Without having dived into the source code (yet), there's a few things that I've learned just by examining the files themselves.

 - Looking at one file, CTOPFixed.DCF, the number of columns in a given row varies from 5 columns in one row to 40 columns in another row.
![image.png](https://images.zenhubusercontent.com/626ab5e81394ace64f8ed65d/5f099830-7d18-431d-b1f9-eaf03e9fc984)

 - Each line starts with a short value that designates what type of object it refers to.  "N" seems to refer to contacts with the seventh column then containing the entity type.
 ![image.png](https://images.zenhubusercontent.com/626ab5e81394ace64f8ed65d/82db1fc2-a921-496a-b6be-dc83e47e107b)

 - The second column in every row is a numeric value in sequential order that serves as an ID.  ID's for contacts are included in other rows (which are currently presumed to be transactions)
 
 - The ID values found in the second column are never duplicated in any of the example files.  In _one_ file, the ID's are not in a rising sequential order but instead dip to lower values for brief windows before returning to the proper sequential order.
 
[The example DCf files](https://app.zenhub.com/files/446466944/aab63e6f-0398-4c9f-ab87-9245705e4aeb/download)

Elaine-Krauss-TCG:
The example files were previously included, but on second thoughts, they might include privileged information...  I can provide them in a non-open source way to any team members who are interested.

Elaine-Krauss-TCG:
N-type entries (Contacts) have been fully documented as recorder [here](https://docs.google.com/spreadsheets/d/1Wyq2Q5X7got_OLw1lwMecTK6SP0BNMBFORnUFrVXrh4/edit#gid=0)

Elaine-Krauss-TCG:
[...]

PR #443, opened 2023-05-23
Feature/400 - DCF Importing Research

Elaine-Krauss-TCG:
#400

Creates an app that will eventually contain the dcf file importing app, and includes the script that I wrote to help parse and poke at DCF files

This is included only because code snippets from it _might_ be helpful later down the line


django-backend/fecfiler/dcf_importer/dcf_parser.py
```diff
@@ -0,0 +1,145 @@
...
+args = parser.parse_args()
+VERBOSE = args.verbose  # Controls whether or not checks for minor errors are run
+DEBUG = args.debug      # If true, script will print its state in great detail as it runs
+DCF_FILENAME = args.dcf_filename
+ROW_TYPE = args.row_type
+COL_NUMBERS = args.col_numbers
+
+
+# Merges extra columns that were made when a comma was included in a quotation
+def get_sanitized_row(line, row_number):
```
toddlees, line 40: found this cool implementation for people running into this problem
https://stackoverflow.com/a/21527375

Basically, python has a csv utility and you can tell it what the delimiter is and what kinds of quotes supersede the delimiter.

No need to change, but I thought you'd be interested

toddlees (approved):

django-backend/fecfiler/dcf_importer/dcf_parser.py
Elaine-Krauss-TCG: Oh sick, I will keep that in mind for the future!  We will probably want to use that method when building out the actual importing, just to better cover any potential edge cases.

PR #447, opened 2023-06-05
Feature/642,739 Create in-kind receipt and In-kind out transactions

mjtravers:
https://github.com/fecgov/fecfile-web-app/issues/642
https://github.com/fecgov/fecfile-web-app/issues/739


django-backend/fecfiler/transactions/schedule_b/managers.py
```diff
@@ -28,6 +28,7 @@
     "NATIONAL_PARTY_RECOUNT_ACCOUNT_DISBURSEMENT",
     "NATIONAL_PARTY_HEADQUARTERS_ACCOUNT_DISBURSEMENT",
     "NATIONAL_PARTY_CONVENTION_ACCOUNT_DISBURSEMENT",
+    "IN_KIND_OUT",
```
toddlees, line 31: IN_KIND_OUT is always itemized regardless of amount according to the ticket, so it should not be in this list.

toddlees (changes requested):

django-backend/fecfiler/transactions/schedule_b/managers.py
mjtravers: This requirement changed recently and the acceptance criteria is to auto itemize only  for > $200.01

toddlees (approved):

PR #481, opened 2023-07-05
feature/1000,1002 Add individual loan forms

mjtravers:


django-backend/fecfiler/transactions/fixtures/payload_transaction_pair.json
```diff
@@ -0,0 +1,135 @@
+{
```
toddlees, line 1: I'm a little hesitant to put a data file that's not an actual django fixture in the fixures folder.  I would put it in `tests/data/...` but i think that's just a personal preference

toddlees (approved):

PR #530, opened 2023-09-22
Feature/1069 Add loan pull forward feature

mjtravers:
Add loan pull forward from previous report.


django-backend/fecfiler/f3x_summaries/models.py
```diff
@@ -368,6 +372,72 @@ class F3XSummary(SoftDeleteModel, CommitteeOwnedModel):
...
+                self.pull_forward_loans()
+
+    def pull_forward_loans(self):
+
+        previous_report = (
+            F3XSummary.objects.get_queryset()
+            .filter(
+                committee_account=self.committee_account,
+                coverage_through_date__lt=self.coverage_through_date,
+                deleted__isnull=True,
```
toddlees, line 394: `deleted__isnull=True,` should be implicit if using `F3XSummary.objects` since it inherits softdelete model

django-backend/fecfiler/f3x_summaries/models.py
```diff
@@ -368,6 +372,72 @@ class F3XSummary(SoftDeleteModel, CommitteeOwnedModel):
...
+            .last()
+        )
+        if previous_report:
+            loans_to_pull_forward = F3XSummary.objects.get(
+                id=previous_report.id
+            ).transaction_set.filter(
+                ~Q(loan_balance=Decimal(0.0)) | Q(loan_balance=None),
+                ~Q(memo_code=True),
+                schedule_c_id__isnull=False,
+                deleted__isnull=True,
```
toddlees, line 406: same as above comment

django-backend/fecfiler/f3x_summaries/models.py
```diff
@@ -368,6 +372,72 @@ class F3XSummary(SoftDeleteModel, CommitteeOwnedModel):
...
+                coverage_through_date__lt=self.coverage_through_date,
+                deleted__isnull=True,
+            )
+            .order_by("coverage_through_date")
+            .last()
+        )
+        if previous_report:
+            loans_to_pull_forward = F3XSummary.objects.get(
+                id=previous_report.id
+            ).transaction_set.filter(
```
toddlees, line 402: I believe `previous_report` and `F3XSummary.objects.get(
                id=previous_report.id
            )` are equivalent.
 can this be `previous_report.transaction_set.filter(...`

django-backend/fecfiler/f3x_summaries/models.py
```diff
@@ -368,6 +372,72 @@ class F3XSummary(SoftDeleteModel, CommitteeOwnedModel):
...
+                schedule_c_id__isnull=False,
+                deleted__isnull=True,
+            )
+
+            for loan in loans_to_pull_forward:
+                # Save children as they are lost from the loan object
+                # when the loan is saved
+                loan_children = copy.deepcopy(loan.children)
+
+                loan.schedule_c.pk = loan.schedule_c.id = None
```
toddlees, line 414: https://docs.djangoproject.com/en/4.2/topics/db/queries/#copying-model-instances
to follow django docs, we probably want to also set `loan._state.adding = True`

django-backend/fecfiler/f3x_summaries/models.py
```diff
@@ -368,6 +372,72 @@ class F3XSummary(SoftDeleteModel, CommitteeOwnedModel):
...
+                loan.schedule_c.save()
+                loan.schedule_c_id = loan.schedule_c.id
+                loan.report_id = self.id
+                loan.report = self
+                # The loan_id should point to the original loan transaction
+                # even if the loan is pulled forward multiple times.
+                loan.loan_id = loan.loan_id if loan.loan_id else loan.id
+                loan.loan = (
```
toddlees, line 422: I think this line is unnecessary after setting the loan_id

toddlees (changes requested):

django-backend/fecfiler/f3x_summaries/models.py
mjtravers: change made

django-backend/fecfiler/f3x_summaries/models.py
mjtravers: change made

django-backend/fecfiler/f3x_summaries/models.py
mjtravers: change made

django-backend/fecfiler/f3x_summaries/models.py
mjtravers: change made

django-backend/fecfiler/f3x_summaries/models.py
mjtravers: change made

django-backend/fecfiler/transactions/migrations/0022_add_debt_foreign_key.py
```diff
@@ -17,7 +17,7 @@ class Migration(migrations.Migration):
                 blank=True,
                 null=True,
                 on_delete=django.db.models.deletion.CASCADE,
-                related_name="debt_repayments",
```
toddlees, line 20: i think this might cause problems on DEV since this migration already ran on it

toddlees: You'll need to revert this change and add the related_name change to a later migration

django-backend/fecfiler/f3x_summaries/models.py
```diff
@@ -368,9 +372,72 @@ class F3XSummary(SoftDeleteModel, CommitteeOwnedModel):
...
+                loan_children = copy.deepcopy(loan.children)
+
+                loan.schedule_c_id = self.copy_fkey(loan.schedule_c)
+                loan.memo_text_id = self.copy_fkey(loan.memo_text)
+                loan.report_id = self.id
+                loan.report = self
+                # The loan_id should point to the original loan transaction
+                # even if the loan is pulled forward multiple times.
+                loan.loan_id = loan.loan_id if loan.loan_id else loan.id
+                self.copy_fkey(loan)
```
toddlees, line 417: the name `copy_fkey` doesn't really seem to describe what's happening here. what do you think about naming it `copy_record`?

toddlees (changes requested):

django-backend/fecfiler/transactions/migrations/0022_add_debt_foreign_key.py
toddlees: Actually! according to SO, related_name is not state on the DB so it doesn't matter where it shows up in the migration https://stackoverflow.com/a/19815633

django-backend/fecfiler/f3x_summaries/models.py
mjtravers: yes, that's  better

toddlees (approved):

PR #535, opened 2023-09-27
Feature/798 Add force unaggregate to schedule A

mjtravers:
#798


django-backend/fecfiler/transactions/managers.py
```diff
@@ -64,9 +64,18 @@ def get_queryset(self):
             date=OuterRef("date"), created__lte=OuterRef("created")
         )
         group_clause = Q(aggregation_group=OuterRef("aggregation_group"))
+        force_unaggregated_clause = Q(force_unaggregated=False) | Q(
```
toddlees, line 67: could be ~Q(force_unaggregated=True)

toddlees (approved):

django-backend/fecfiler/transactions/models.py
```diff
@@ -68,6 +68,7 @@ def form_type(self, value):
     entity_type = models.TextField(null=True, blank=True)
     memo_code = models.BooleanField(null=True, blank=True, default=False)
     force_itemized = models.BooleanField(null=True, blank=True)
```
toddlees, line 70: think you need to add force_itemized to the transaction serializer.  it's not getting saved when we toggle it on the front end

toddlees (changes requested):

django-backend/fecfiler/transactions/managers.py
mjtravers: When working on the loan/debt pull forward, I found that ~Q(some_field=True) did not include the records with NULL, just the records with False. Do you not find this to not be the case? I will double-check.

django-backend/fecfiler/transactions/models.py
mjtravers: I thought about that but then I was able to save without any issues on my local. Were you not able to update locally? I can retry and verify.

toddlees: Not needed because we don't do anything fancy with it in the serializer so it gets pulled in with the .get_fields

django-backend/fecfiler/transactions/managers.py
mjtravers: Made this correction and updated pull forward query as well.

toddlees (approved):

PR #544, opened 2023-10-10
feature/502 - remove scha_transactions django app

dheitzer:
Note that the **0002_copy_scha.py** migration operations is left to [] which means nothing happens for this migration since it referenced the scha_transactions model and copying is no longer needed.


django-backend/fecfiler/transactions/migrations/0002_copy_scha.py
```diff
@@ -1,37 +1,6 @@
 # Generated by Django 3.2.12 on 2022-12-13 22:39
 
 from django.db import migrations
```
Elaine-Krauss-TCG, line 3: We _might_ want to leave a comment for later explaining why this is empty.  It's not essential, though.

Elaine-Krauss-TCG (approved):

toddlees (approved):

PR #545, opened 2023-10-11
Fix manage transactions table sorting

mjtravers:


django-backend/fecfiler/transactions/views.py
```diff
@@ -106,45 +106,62 @@ class TransactionListPagination(pagination.PageNumberPagination):
...
+                    "schedule_b__payee_first_name",
+                    "schedule_c__lender_first_name",
+                    "schedule_d__creditor_first_name",
+                    "schedule_e__payee_first_name",
                 ),
                 output_field=TextField(),
             ),
-        )
+        ),
+        sort_amount=Coalesce(
```
toddlees, line 138: `amount` and `date` are already working like this in the Transaction manager.  they're just missing schedule e

toddlees (changes requested):

toddlees (approved):

PR #547, opened 2023-10-11
Fix election_code pattern for candidate_state validation

mjtravers:


mjtravers:
OBE. Closing without merging.

PR #610, opened 2023-11-16
Patch 1182

toddlees:
What do you think of this? my goal is to avoid any possibility of error when implementing the lines comes around.  I think this enforces that when we implement the lines, we must get rid of the scaffolding, and there is no way to do it without


Elaine-Krauss-TCG:
I see what you're getting at, but personally I lean a bit against this PR.  If the scaffolding as I've implemented it isn't removed, new unit tests will fail.  Also, the comments make it very obvious where things have been stubbed out (though the comments could be added to your PR, of course).  And for one more thing, _imo_ ` = Decimal(0)` has much clearer behavior than `get_line(None)`

I'm certainly not *strongly* against these changes, though.

PR #616, opened 2023-11-28
feature/1543

dheitzer:
[#1543](https://app.zenhub.com/workspaces/fecfile-online-619e578e68408b001c831251/issues/gh/fecgov/fecfile-web-app/1543)

we're just going to use a numerical sort order to address this

### DIFF
--- a/django-backend/fecfiler/transactions/managers.py
+++ b/django-backend/fecfiler/transactions/managers.py
@@ -269,8 +269,10 @@ class TransactionManager(SoftDeleteManager):
                 line_label=Case(
                     # Schedule A
                     When(_form_type="SA11A", then=Value("11(a)")),
-                    When(_form_type="SA11AI", then=Value("11(a)(i)")),
-                    When(_form_type="SA11AII", then=Value("11(a)(i)")),
+                    When(_form_type="SA11AI", itemized=True, then=Value("11(a)(i)")),
+                    When(_form_type="SA11AI", itemized=False, then=Value("11(a)(ii)")),
+                    When(_form_type="SA11AII", itemized=False, then=Value("11(a)(ii)")),
+                    When(_form_type="SA11AII", itemized=True, then=Value("11(a)(i)")),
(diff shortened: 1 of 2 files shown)


django-backend/fecfiler/transactions/managers.py
```diff
@@ -269,8 +269,10 @@ def get_queryset(self):
                 line_label=Case(
                     # Schedule A
                     When(_form_type="SA11A", then=Value("11(a)")),
-                    When(_form_type="SA11AI", then=Value("11(a)(i)")),
-                    When(_form_type="SA11AII", then=Value("11(a)(i)")),
+                    When(_form_type="SA11AI", itemized=True, then=Value("11(a)(i)")),
```
toddlees, line 272: i wonder if you moved the line_label clause to a later annotate, would you be able to use `form_type` rather than `_form_type`.  `form_type` should be mapped correctly based on itemization

toddlees (approved):

django-backend/fecfiler/transactions/managers.py
dheitzer: > i wonder if you moved the line_label clause to a later annotate, would you be able to use `form_type` rather than `_form_type`. `form_type` should be mapped correctly based on itemization

I tried to do this when first implementing the line_label case and it works, however it adds massively to the query size (since it generates large subqueries for calculating form_type for each case).  Also, the query performance degraded noticeably when I did this also.

dheitzer: > I also wonder, we put `line_label` in the api because we needed to sort on it. Since we're sorting on `line_label_order_key` now, should the `line_label` mapping be defined on the front end now? I'm not sure.

I lean towards leaving it in the api since it is coupled to the line_label_order_key definition and adding a new label in app would mean still adding a line_label_order_key in api (which could more easily be forgotten I feel if they're not in the same place).  I am open if we want to move it however.

mjtravers (approved):

PR #628, opened 2023-12-13
216 - Add endpoint to fetch candidate details so the front end doesn't need to know the fec api key

sasha-dresden:


sasha-dresden:
Sorry for all the pushes. Was struggling to understand the sonarcloud security and linting. One issue it didn't like was using user input as part of the url, which is required in order to hit the fec api. I read up on someone having a similar issue [here](https://community.sonarsource.com/t/change-this-code-to-not-construct-the-url-from-user-controlled-data/34393/6) and according to them we should validate and sanitize. I added some logic for that to get passed the sonar test, but we might want to make it a bit more secure.
(To be fair, according to the sonar sourcer this is a known issue with their process as adding user input on to the tail end of a url shouldn't trigger the same failure as if they could alter the entire url, specifically the domain)

django-backend/fecfiler/contacts/views.py
```diff
@@ -32,6 +33,14 @@
 )
 
 
+def validate_candidate(candidate_id):
+    return len(candidate_id) == 9
+
+
+def sanitize(candidate_id):
+    return candidate_id
```
mjtravers, line 41: Perhaps to sanitize, we could check for valid characters. I believe it's just numbers and a handful of capital letters. Let's check with Gregg.

mjtravers (changes requested):

mjtravers (approved):

PR #652, opened 2024-01-04
Feature/592 - database baseline (DO NOT MERGE)

dheitzer:
[Create a new Django migration baseline on PROD based on current models and Postgres upgrade](https://app.zenhub.com/workspaces/fecfile-online-619e578e68408b001c831251/issues/gh/fecgov/fecfile-web-api/592)


django-backend/fecfiler/reports/migrations/0002_initial.py
```diff
@@ -1,24 +1,20 @@
-# Generated by Django 3.2.12 on 2022-09-15 17:09
```
mjtravers, line 1: Can this migration file be squashed into the 0001_initial.py file?

mjtravers (commented):
See inline comment about the 2 migration files for the report app.

django-backend/fecfiler/reports/migrations/0002_initial.py
dheitzer: @mjtravers this particular one can't be squashed since reports has a circular dependency with web_service and the models might need reworked to avoid this.  I was however able to squash the memo_text 0002 migration that got generated for some reason.

mjtravers: Sounds good. Thanks

mjtravers (approved):

PR #673, opened 2024-01-18
1555 - Form 1M qualifications

mjtravers:
Issue: https://github.com/fecgov/fecfile-web-app/issues/1555

Add qualifications candidates to the F1M entry form.


django-backend/fecfiler/contacts/views.py
```diff
@@ -39,6 +38,28 @@ def validate_and_sanitize_candidate(candidate_id):
     return candidate_id
 
 
+def save_payload_contacts(payload, contact_keys, request):
```
toddlees, line 41: I'm seeing overlap between `save_payload_contacts` and `create_or_update_contact`.  Seems like `write_contacts` =  `create_or_update_contact` X `contact_keys` = `save_payload_contacts` but in the serializer instead of the view.  I wonder if we can share the code

toddlees (changes requested):

Elaine-Krauss-TCG (approved):

dheitzer (approved):

PR #681, opened 2024-01-23
Give test committee a uuid

toddlees:
If you try to start docker up again locally, the e2e-test-data.json fails to load because the C00601211 committee colides with itself.  Giving it an id changes it to just write to the same record.

NOTE:  You certainly have a different uuid for that committee locally.  If you need to keep your local data for some reason, you could change the pk in the fixture to match yours until you're ready for the clean slate.

### DIFF
--- a/django-backend/fixtures/e2e-test-data.json
+++ b/django-backend/fixtures/e2e-test-data.json
@@ -1,72 +1,73 @@
 [
-{
-    "model": "authentication.account",
-    "pk": 165,
-    "fields": {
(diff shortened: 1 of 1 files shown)


mjtravers (approved):

lbeaufort:
Noting that because we run `python manage.py loaddata fixtures/e2e-test-data.json` every time we restart the app, we'll see something like:
```2024-01-23 09:30:34 2024-01-23T14:30:34.767338Z [info     ] Committee Account: C00601211 was updated [fecfiler.committee_accounts.signals]```
in the logs, and that's expected behavior because of the way `loaddata` works

PR #690, opened 2024-01-26
1598 Use membership model to represent user's relation to committees

toddlees:
New, much simplified user model.  The model before handrolled almost everything that was already implemented by django.

Creates Membership model which is a join table between users and committees.


django-backend/fecfiler/committee_accounts/views.py
```diff
@@ -1,15 +1,58 @@
...
+            return self.get_paginated_response(serializer.data)
+
+        serializer = CommitteeMemberSerializer(
+            queryset, many=True, context=serializer_context
+        )
+        return Response(serializer.data)
+
+    @action(detail=True, methods=["post"])
+    def activate_committee(self, request):
+        committee = self.get_object()
```
dheitzer, line 38: probably obvious but we'll need to add security checks here to make sure a user has access to the committee account

django-backend/fecfiler/user/models.py
```diff
@@ -0,0 +1,20 @@
...
+class User(AbstractUser):
+    id = models.UUIDField(
+        default=uuid.uuid4,
+        editable=False,
+        primary_key=True,
+        serialize=False,
+        unique=True,
+    )
+    first_name = models.CharField(
+        _("first name"), max_length=150, null=True, blank=True
```
dheitzer, line 16: small nit: we probably don't need _("first name") and _("last name")

dheitzer (approved):
looks nice

django-backend/fecfiler/committee_accounts/views.py
toddlees: yeah the queryset handles that because it will only return committees that the user is part of

django-backend/fecfiler/user/models.py
toddlees: agreed.  pulled it from django source because we wanted to add `null=True`

mjtravers (approved):

PR #727, opened 2024-02-20
Feature/91 - Create new endpoint to remove a user from a committee

sasha-dresden:
Includes a front end PR that goes along with this.
There was an odd issue with the sonarcloud being unhappy about the safe/unsafe http methods. I spoke to @mjtravers about this and he agreed this is probably something we can safely ignore as it's warning that this method is unsafe (meaning it's a DELETE, POST, or PUT request), which is what it needs to be in order to change the system.


sasha-dresden:
Updated to match with updates for adding a user to a committee.

Elaine-Krauss-TCG (approved):

mjtravers (approved):

PR #759, opened 2024-02-29
689 - Update output to print to sysout while refusing all errors or exceptions that are not validation related from going to the front end

sasha-dresden:


django-backend/fecfiler/utils.py
```diff
@@ -19,8 +19,11 @@ def custom_exception_handler(exc, context):
     if response is not None and response.status_code == 403:
         delete_user_logged_in_cookies(response)
 
-    # Do not allow an error response body
-    # if getattr(response, 'data'):
-    #     response.data = None
+    # Do not allow an error response body unless validation
+    data = getattr(response, 'data')
+    exception_type = type(exc)
+    print(data)
```
lbeaufort, line 25: @sasha-dresden I think we should avoid using `print` statements

lbeaufort (changes requested):
I think we should avoid using `print` statements

django-backend/fecfiler/utils.py
sasha-dresden: Good point. Switched it over to use structlog

django-backend/fecfiler/utils.py
```diff
@@ -1,26 +1,32 @@
...
         delete_user_logged_in_cookies(response)
 
-    # Do not allow an error response body
-    # if getattr(response, 'data'):
-    #     response.data = None
+    # Do not allow an error response body unless validation
+    data = getattr(response, 'data')
+    exception_type = type(exc)
+    logger.error("Error: {}".format(data))
+    if data and "ValidationError" not in str(exception_type):
```
toddlees, line 29: Can avoid string parsing by using `exception_type is not ValidationError`

importing:
`from rest_framework.exceptions import ValidationError`

django-backend/fecfiler/utils.py
```diff
@@ -1,26 +1,32 @@
...
+    if response.status_code == 403:
         delete_user_logged_in_cookies(response)
 
-    # Do not allow an error response body
-    # if getattr(response, 'data'):
-    #     response.data = None
+    # Do not allow an error response body unless validation
+    data = getattr(response, 'data')
+    exception_type = type(exc)
+    logger.error("Error: {}".format(data))
```
toddlees, line 28: 3.6 introduced f-strings which tend to be more readable https://peps.python.org/pep-0498/
`logger.error(f"Error: {data}")`

toddlees (changes requested):

django-backend/fecfiler/utils.py
sasha-dresden: Ah that was a simple enough update and a lot cleaner

django-backend/fecfiler/utils.py
sasha-dresden: Oh I think I tried that previously but I imported the wrong ValidationError, I think I tried importing from django.forms and that didn't seem to work so I just used string comparison. But class validation is much better. Got that updated with the correct ValidationError now.

toddlees (approved):

PR #760, opened 2024-03-01
66 - Add endpoint to get transactions for a contact

sasha-dresden:
There is a web app PR that goes along with this.

### DIFF
--- a/django-backend/fecfiler/contacts/views.py
+++ b/django-backend/fecfiler/contacts/views.py
@@ -1,3 +1,4 @@
+from uuid import UUID
 import structlog
 
 import re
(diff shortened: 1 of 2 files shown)


django-backend/fecfiler/contacts/views.py
```diff
@@ -264,6 +267,16 @@ def get_contact_id(self, request):
         )
         return Response(match.id if match else "")
 
+    @action(detail=True, url_path="transactions")
+    def get_contact_transactions(self, request, pk: UUID):
```
toddlees, line 271: The front end table is expecting a paginated response.  There is a future requirement to filter transactions by contact in the 'all transactions' view.  You can solve this need and that by adding a contact query filter to the transaction list view instead of a new endpoint.  We've done this with `parent_id` in /transactions/views.py list().  
```
parent_id = self.request.query_params.get("parent")
        if parent_id:
            queryset = queryset.filter(parent_transaction_id=parent_id)
```

you could do something similar with `contact_id`. 

This would also simplify the front end code because you could use the table values like `amount` and `date` instead of having to define `getTransactionDate()` and `getTransactionAmount()` on the front ennd

toddlees (changes requested):

django-backend/fecfiler/contacts/views.py
sasha-dresden: Ah I didn't know about the future requirement so I didn't think a paginated response would be necessary. Switching it over to use the transactions/views.py with contact_id was super easy and worked perfectly. Updated both front end and api to work with this new method.

toddlees (approved):

PR #765, opened 2024-03-08
689 - log errors and raise validation error in committee register

toddlees:
We were losing logs for many errors if response came back `None`

Also changes BadRequest to ValidationError in the register committee view so it can get to the front end

### DIFF
--- a/django-backend/fecfiler/committee_accounts/views.py
+++ b/django-backend/fecfiler/committee_accounts/views.py
@@ -201,11 +201,15 @@ def register_committee(committee_id, user):
             f1_line = dot_fec_content.split("\n")[1]
             f1_email = f1_line.split(FS_STR)[11]
 
+    if f1_email != email:
+        raise ValidationError(f"Email {email} does not match committee email")
+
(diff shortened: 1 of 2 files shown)


django-backend/fecfiler/committee_accounts/views.py
```diff
@@ -198,7 +205,7 @@ def register_committee(committee_id, user):
         committee_id=committee_id
     ).first()
     if existing_account or f1_email != email:
-        raise BadRequest("could not register committee")
+        raise ValidationError("could not register committee")
```
lbeaufort, line 208: Would context on why the committee couldn't be registered be useful here? Because it could be there is an `existing_account` or `f1_email != email`

mjtravers (approved):

django-backend/fecfiler/committee_accounts/views.py
mjtravers: The exception message was split up so that a more specific error message is now presented. The email check was moved up so to short circuit the need for a committee check should the email not match.

PR #769, opened 2024-03-11
656 - transaction performance refactoring

mjtravers:
Refactoring of tables to include views for performance improvements.


django-backend/fecfiler/settings/base.py
```diff
@@ -375,6 +377,8 @@ def get_env_logging_processors(log_format=LINE):
 FEC_API_CANDIDATE_LOOKUP_ENDPOINT = str(FEC_API) + "candidates/"
 FEC_API_CANDIDATE_ENDPOINT = str(FEC_API) + "candidate/"
 
+SILKY_PYTHON_PROFILER = True
```
lbeaufort, line 380: @mjtravers I think this PR is for visibility and @toddlees wasn't ready for feedback, but just in case, I think we need to remove this before it's merged, or turn it off in production

toddlees (approved):

PR #785, opened 2024-03-21
Mod/release/sprint 28 conn max age

lbeaufort:
## Summary

Test conn_max_age

>conn_max_age sets the [CONN_MAX_AGE setting](https://docs.djangoproject.com/en/stable/ref/settings/#conn-max-age), which tells Django to persist database connections between requests, up to the given lifetime in seconds. If you do not provide a value, it will follow Django’s default of 0. Setting it is recommended for performance.

https://pypi.org/project/dj-database-url/#:~:text=conn_max_age%20sets%20the%20CONN_MAX_AGE%20setting,it%20is%20recommended%20for%20performance. 

## How to test


lbeaufort:
Wrong branch...

PR #797, opened 2024-03-29
Feature/386 - Add check to see if a report can be deleted

sasha-dresden:


django-backend/fecfiler/reports/views.py
```diff
@@ -177,6 +180,70 @@ def list(self, request, *args, **kwargs):
...
+    @action(
+        detail=False,
+        methods=["get"],
+        url_path="can_delete",
+    )
+    def can_delete(self, request):
+        report_id = (request.query_params.get("report_id")
+                     or request.data.get("report_id"))
+        report_id = report_id.replace("-", "")
+        with connection.cursor() as cursor:
```
toddlees, line 192: We should be able to write these queries with the Django framework.  Although maybe a reason to handwrite it would be to do it in one query.  You could probably do select on transactions, a join on the relation table for reports, and another join on the transaction table where the back_reference_id equals the row's transaction id.  they you filter by rows that were joined with the back_reference_id

toddlees (changes requested):

django-backend/fecfiler/reports/views.py
mjtravers: In addition to Todd's comments, since the delete check is only needed on the Manage Reports page, the query_set could only be modified to add the canDelete flag in an override of the list() DRF view. The SQL joins have been problematic so isolating additional ones where we can is preferred. Also, if performance becomes and issue, coding it so that only transactions who turn up in the result set due to pagination get the delete flag annotation could be a possibility. See this new ticket for how we're looking into making this sort of modification to the Contact Manager which is having issues with query performance stemming from its delete flag: https://github.com/fecgov/fecfile-web-api/issues/799

sasha-dresden: Ok. Thank you both so much for helping me with this. 
I removed this can_delete end point and instead built it in the serializer using django queries.
I also updated the delete functionality for reports and transactions to be hard deletes.
Additionally, apparently django 4.2.11 is now required by circle for dependency testing to pass, so I updated that here and in [the mozilla project.](https://github.com/mozilla/mozilla-django-oidc/pull/534)

django-backend/fecfiler/reports/tests/test_models.py
```diff
@@ -30,3 +30,18 @@ def test_amending_f24(self):
             f24_report.form_24.original_amendment_date, new_upload_submission.created
         )
         self.assertEquals(f24_report.form_type, "F24A")
+
+    def test_delete(self):
+        f24_report = Report.objects.get(id="10000f24-d926-4e89-ad4b-000000000001")
+        f24_report.delete()
+
```
mjtravers, line 37: Can we add assertions to test that transactions are not soft deleted for an F24 and are soft deleted otherwise?

mjtravers (changes requested):
See inline comment about adding a couple of asserts to the unit test.

toddlees (approved):
Looks great pending matt's test

mjtravers (approved):

PR #802, opened 2024-04-03
Feature/1755 - feedback dialog

dheitzer:
## Summary

## How to test

-

### DIFF
--- a/django-backend/fecfiler/feedback/views.py
+++ b/django-backend/fecfiler/feedback/views.py
@@ -20,7 +20,7 @@ class FeedbackViewSet(viewsets.ViewSet):
         serializer = FeedbackSerializer(data=request.data, context={"request": request})
         if not serializer.is_valid():
             return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
-        title = "User feedback on " + request.META.get("Referer")
+        title = "User feedback on " + serializer.validated_data["location"]
 
         body = (
             "## What were you trying to do and how can we improve it?\n %s \n\n"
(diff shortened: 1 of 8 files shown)


Elaine-Krauss-TCG (approved):

django-backend/fecfiler/feedback/views.py
```diff
@@ -0,0 +1,43 @@
...
+        ) % (
+            serializer.validated_data["action"],
+            serializer.validated_data["feedback"],
+            serializer.validated_data["about"],
+            serializer.validated_data["location"],
+            request.META["HTTP_USER_AGENT"],
+        )
+
+        client = github3.login(token=FECFILE_GITHUB_TOKEN)
+        client.repository("fecgov", "fecfile-feedback").create_issue(title, body=body)
```
toddlees, line 41: we should probably handle the scenario where submitting fails, at least to log it.

django-backend/fecfiler/feedback/views.py
```diff
@@ -0,0 +1,43 @@
...
+
+        body = (
+            "## What were you trying to do and how can we improve it?\n %s \n\n"
+            "## General feedback?\n %s \n\n"
+            "## Tell us about yourself\n %s \n\n"
+            "## Details\n"
+            "* URL: %s \n"
+            "* User Agent: %s"
+        ) % (
+            serializer.validated_data["action"],
```
toddlees, line 33: Does github sanitize content to it's api?  wondering if we're responsible for removing dangerous content like scripts

toddlees (dismissed):

django-backend/fecfiler/feedback/views.py
dheitzer: added a try/catch to this section to log any errors (and returns a 500)

django-backend/fecfiler/feedback/views.py
dheitzer: great question - I couldn't find a definitive answer for this so I added escaping on our end to the title and body.

Issue #799, opened 2024-04-02
Move transaction and report counts  from Contact query set annotation to list view query.


mjtravers:
### Business Reason ### 

As a user, I will be able to display the contact list page so that I can see it download quickly.

### Acceptance Criteria ### 

**If** viewing the Contact Manager list page
**When** I view the page
**Then** it will not lag when displaying

### DEV Notes ###
This bit of code is adding a number of joins to the queryset and significantly slowing down the display of the contact manager list page:

https://github.com/fecgov/fecfile-web-api/blob/develop/django-backend/fecfiler/contacts/views.py#L58-L70

The transaction_count and report_count annotations are only used in one place in the front end, to set the delete trashcan for a contact to disabled or enabled:

https://github.com/fecgov/fecfile-web-app/blob/develop/front-end/src/app/contacts/contact-list/contact-list.component.ts#L67

**Possible solution:**
Remove the count annotations from the queryset and move the calculations to the Django callback that produces the response to the list API endpoint. The idea is the counts can be limited to just the 5 - 20 records that are produced by the list paginator. Also, we currently display a count but we could change the output to a boolean if that can be created with a faster query.

https://www.django-rest-framework.org/api-guide/viewsets/

Benchmark the api response change before and after the change using the request timer logging.

mjtravers:
Passes CR. Sending to QA.

mjtravers:
Unit tests passing after code changes
https://app.circleci.com/pipelines/github/fecgov/fecfile-web-api/3473/workflows/532efe59-859c-4d42-a79d-4f86a1f8afea/jobs/10510
![Image](https://github.com/fecgov/fecfile-web-api/assets/2971785/6dbbf6a9-5485-4a3f-8d8f-260f47ee23bd)

WiseQA:
QA review verified per DEV unit tests passing.

![Image](https://github.com/fecgov/fecfile-web-api/assets/96139912/890f8f00-2ffe-452a-8edb-1a341222b659)

Moved to Stage Ready.

PR #810, opened 2024-04-09
799 - Speed up check to see if contact has transactions or reports associated with them

sasha-dresden:
Replace transaction_count and report_count with singular boolean exists check and move to serializer so it'll only run for serialized data


Elaine-Krauss-TCG:
I ran a bunch of locust testing on these, and in every run with the 799 changes, I get at least one response for `load_contacts` (which loads one page of 10 contacts) in the neighborhood of 200ms.  That said, the code looks great, and everything _feels_ great running locally.

Locust testing on #799
image

Locust testing on develop
image

mjtravers (approved):

PR #811, opened 2024-04-10
 Update Report::delete to make sure the properties marked as CASCADE are also deleted

sasha-dresden:


django-backend/fecfiler/reports/models.py
```diff
@@ -137,6 +137,16 @@ def delete(self):
...
         ReportTransaction.objects.filter(report=self).delete()
+
+        for field in self._meta.fields:
+            if (
+                isinstance(field, models.ForeignKey)
+                and field.remote_field.on_delete == models.CASCADE
+            ):
+                foreign_key = getattr(self, field.name)
+                if foreign_key:
+                    foreign_key.delete()
```
toddlees, line 148: image

this includes the committee account the report is owned by.

You could loop through TABLE_TO_FROM:

```
for form_key in self.TABLE_TO_FROM:
  form = getattr(self, form_key)
  if form:
     form.delete()
```

toddlees (changes requested):

django-backend/fecfiler/reports/models.py
sasha-dresden: Replaced with iterating through the table as suggested.

mjtravers (approved):

toddlees (approved):

PR #822, opened 2024-04-11
787 - Set `conn_max_age`, update `dj-database-url` minor version

lbeaufort:
## Summary

- Issue api#787
  
 Set `conn_max_age` to 600 seconds (10 min), update `dj-database-url` minor version. Confirmed that the cloud.gov postgres RDS doesn't have a timeout set

From https://pypi.org/project/dj-database-url/:

>`conn_max_age` sets the [CONN_MAX_AGE setting](https://docs.djangoproject.com/en/stable/ref/settings/#conn-max-age), which tells Django to persist database connections between requests, up to the given lifetime in seconds. If you do not provide a value, it will follow Django’s default of 0. Setting it is recommended for performance.

>`conn_health_checks` sets the [CONN_HEALTH_CHECKS setting](https://docs.djangoproject.com/en/stable/ref/settings/#conn-health-checks) (new in Django 4.1), which tells Django to check a persisted connection still works at the start of each request. If you do not provide a value, it will follow Django’s default of False. Enabling it is recommended if you set a non-zero conn_max_age.

## How to test

- After pulling this branch and restarting the API container, run:
 `docker exec -it fecfile-api sh -c "python manage.py diffsettings --all" | grep DATABASES` 
to see the `CONN_MAX_AGE` database setting. 
- Can load test with locust before/after. My results: https://docs.google.com/spreadsheets/d/17m9kO1fkfxRNlgAm3u916BzmeE0n82lKLkUUNvbwmYA/edit#gid=196778165
- Locust tests show consistent performance improvement in the 50%-80% range of response times. Above 80% varies enough that it's probably symptomatic of the larger query optimization issues covered in https://github.com/fecgov/fecfile-web-api/issues/828


mjtravers (approved):

mjtravers:
Thanks, great write up and capturing of baseline metrics.

PR #824, opened 2024-04-11
hotfix/386-2: Move report_status to serializer so it can be used by can_delete to check if status is In progress

sasha-dresden:
Issue was that I had originally been doing the status check on the front end and then was calling an end point to check if it could be deleted if the status was In progress. However we refactored the can_delete on the api to be part of the serializer, but it was never accounting for the status as it was happening on the front end. But it no longer made sense to have it on the front end since we're just checking against a single can_delete variable which should be accounting for status.

However, to update the can_delete function in the serializer, I needed to move the get_status_mapping() annotation from the view into the serializer, so that the report status would be available to check for the can_delete function.

### DIFF
--- a/django-backend/fecfiler/reports/form_3x/tests/test_serializers.py
+++ b/django-backend/fecfiler/reports/form_3x/tests/test_serializers.py
@@ -5,12 +5,16 @@ from ..serializers import (
 )
 from fecfiler.user.models import User
 from rest_framework.request import Request, HttpRequest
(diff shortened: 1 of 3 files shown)


mjtravers (approved):

mjtravers (changes requested):
The get_status_mapping() needs some additional code coverage.

mjtravers (approved):

PR #835, opened 2024-04-18
Feature/1849 - Add committee_name to Form_3X for missing print preview info

sasha-dresden:
Added committee_name as a column for table report_form3x.
Additionally, on the front end added some code to make sure committee information is sent along with the create form 3X request. [PR1860](https://github.com/fecgov/fecfile-web-app/pull/1860)

I also updated the way downloading a fec file works for local environment. The fec file was being stored in the celery docker container, so in order to retrieve it locally, it needed a celery task, otherwise it was trying to retrieve it from the api container.


django-backend/fecfiler/reports/form_3x/models.py
```diff
@@ -16,6 +16,7 @@ class Form3X(models.Model):
         unique=True,
     )
     change_of_address = models.BooleanField(default=False, null=True, blank=True)
+    committee_name = models.TextField(null=True, blank=True)
```
toddlees, line 19: I think we'll need committee_name in the report model so that the other forms have it

toddlees (changes requested):

django-backend/fecfiler/reports/form_3x/models.py
sasha-dresden: Added committee_name for form 24. We already had it form form 1M and 99, though it does feel like we should move it to a parent class for all forms, but there currently isn't one.

django-backend/fecfiler/web_services/views.py
```diff
@@ -60,7 +62,10 @@ def get_dot_fec(self, request, report_id):
             logger.error(not_found_msg)
             return Response(not_found_msg, status=status.HTTP_400_BAD_REQUEST)
         file_name = dot_fec_record.first().file_name
-        file = get_file(file_name)
+        task = get_dot_fec.apply_async(args=[file_name], retry=False)
```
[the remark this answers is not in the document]
sasha-dresden: Ah, I figured out the problem. When I originally configured my setup, I set my API and DB dockerfiles to the e2e versions, but the Worker was using the default. So it was saving files to nxg_fec but the API was trying to retrieve from nxg_fec_e2e. I fixed my local setup and rolled back those changes.

django-backend/fecfiler/reports/form_3x/models.py
mjtravers: @sasha-dresden It looks like all the current forms have committee_name, street_1, street_2, city, state, and zip. Let's move all of them out of the reports_form* tables and into the reports_report table.

mjtravers:
This PR has been replaced by #839 
Closing without merging.

PR #839, opened 2024-04-19
feature/1849 - Move committee info from forms to report

sasha-dresden:
Issue [APP#1849](https://github.com/fecgov/fecfile-web-app/issues/1849)

This is a replacement for [PR835](https://github.com/fecgov/fecfile-web-api/pull/835) where instead of adding committee_name to Form 3X and Form 24, we move the committee information from the forms into the report table.


toddlees (approved):
looks perfect

Issue #865, opened 2024-04-30
Limit return payload for transaction create and update


mjtravers:
Speed up single transaction create/update response time by not sending back updated transaction to the front end but rather just the transaction id.

![image](https://github.com/fecgov/fecfile-web-api/assets/2971785/3ef1bbfa-c1e6-4aae-9297-197c8c7b3b6b)

toddlees:
passes CR moving to QA

WiseQA:
Per DEV QA verified transactions can be created and updated and the create/edit page navigates to the transactions list page after submitted.

QA Review Completed.   Moved to Stage Ready.

PR #871, opened 2024-05-01
Feature/865 - Make transaction create/update return id

sasha-dresden:
Issue #865 
APP PR [PR1912](https://github.com/fecgov/fecfile-web-app/pull/1912)

### DIFF
--- a/django-backend/fecfiler/transactions/views.py
+++ b/django-backend/fecfiler/transactions/views.py
@@ -81,9 +81,7 @@ class TransactionViewSet(CommitteeOwnedViewMixin, ModelViewSet):
 
         schedule_filters = self.request.query_params.get("schedules")
         if schedule_filters is not None:
-            schedules_to_include = (
-                schedule_filters.split(",") if schedule_filters else []
-            )
+            schedules_to_include = schedule_filters.split(",") if schedule_filters else []
             queryset = queryset.filter(
                 schedule__in=[
                     Schedule[schedule].value for schedule in schedules_to_include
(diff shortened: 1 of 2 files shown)


django-backend/fecfiler/transactions/views.py
```diff
@@ -108,12 +106,12 @@ def create(self, request, *args, **kwargs):
             saved_transaction = self.save_transaction(request.data, request)
             print(f"transaction ID: {saved_transaction.id}")
         # transaction_view = self.get_queryset().get(id=saved_transaction.id)
-        return Response(TransactionSerializer().to_representation(saved_transaction))
+        return Response(saved_transaction.id)
 
     def update(self, request, *args, **kwargs):
```
toddlees, line 111: we'll want to change `partial_update`, `save_transactions`, and `save_reatt_redes_transactions` to just return ids.

They all call `save_transaction()` which has this return statement: 
```
return self.queryset.get(id=transaction_instance.id)
```
It runs the query again so it can get the correct just-in-time values.  Since we're not going to return the transaction in the payload, what do you think of changing `save_transaction` to return just the id?

toddlees (changes requested):

sasha-dresden:
Updated so save_transactions() returns a list of ids now.
I'm not sure we should update save_transaction() to return just the id, because save_reatt_redes_transactions() needs the transaction created in order to set the reatt_redes property to the newly created transaction

sasha-dresden:
After discussion, just removed save_transactions() as it is no longer used.

toddlees (approved):

Issue #843, opened 2024-04-23
Create the amount calculation fields in the transaction table


mjtravers:
The aggregate calculations are currently being performed just-in-time during the transaction record READ operation.

This spike is to move the calculation from the view to the CREATE, UPDATE, and DELETE operations and save the result in the transactions_transaction table and see how it affects performance for the 3 calculations currently storing their results in the transaction view:

*   entity aggregate
*   calendar year to date per election office
*   loan payment to date

[![](https://github.com/fecgov/fecfile-web-api/assets/110493670/17d8c664-a7c9-4be6-afc3-e97486fcca49)](https://github.com/fecgov/fecfile-web-api/assets/110493670/17d8c664-a7c9-4be6-afc3-e97486fcca49)

Since updating a transaction potentially impacts the aggregate of any transaction in it's topology/network, we need to do write an update that calculates the aggregates and updates the transactions. this should be doable with a single bulk update query that calculates the aggregates with a window function.

### QA Notes ###

 null

### DEV Notes ###

 
 null

### Design ###

 
 null

[FECFILE-181](https://fecgov.atlassian.net/browse/FECFILE-181)

mjtravers:
From CR:

The new aggregation code needs to take into account when a transaction record is edited so that it is removed from a calculation group. For example, if there is a group of 3 transactions for an entity aggregation and one of the transactions has its contact changed, only the updated transaction in its new group is aggregated. The other 2 transactions that were part of the original group are do not have their new aggregation sums without the transaction that was removed from the group recalculated.

To account for this scenario, the trigger code should factor out the window function rollups into separate functions. Then the trigger can call them once for the NEW row record and, if the record has changed so that it is in a new group, then the rollup function can be called on the OLD row record values to update the old group transactions. In postres triggers, NEW and OLD are provided to the function as part of the postgres API.

Also, a couple of unit tests are broken and need to be fixed:
```
======================================================================
FAIL: test_serialize_f3x_summary_instance (fecfiler.web_services.dot_fec.test_dot_fec_serializer.DotFECSerializerTestCase)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/opt/nxg_fec_e2e/fecfiler/web_services/dot_fec/test_dot_fec_serializer.py", line 173, in test_serialize_f3x_summary_instance
    self.assertEqual(split_row[0], "F3XN")
AssertionError: '' != 'F3XN'
+ F3XN

======================================================================
[...]

PR #897, opened 2024-05-24
843 - Migrate calculated fields from on read to on write of transaction

mjtravers:
API #843


django-backend/fecfiler/transactions/calculations.py
```diff
@@ -0,0 +1,156 @@
+# """Calculations for transaction aggregated fields.
```
dheitzer, line 1: confirmed with @sasha-dresden that this can be removed

django-backend/fecfiler/transactions/migrations/0007_transaction__calendar_ytd_per_election_office_and_more.py
```diff
@@ -0,0 +1,172 @@
...
+            ),
+        ),
+        migrations.AddField(
+            model_name="transaction",
+            name="loan_payment_to_date",
+            field=models.DecimalField(
+                blank=True, decimal_places=2, max_digits=11, null=True
+            ),
+        ),
+        migrations.RunSQL(
```
dheitzer, line 34: Just an idea - I wonder if it would be worthwhile and possible to break this up into several smaller functions (similar to calculations.py) for each of the schedules.

sasha-dresden:
Updated aggregate calculator to handle changes when a transaction contact is changed.
Also updated some unit tests to work that were failing due to the unit test rework.

sasha-dresden (approved):

PR #904, opened 2024-06-04
feature/1978 - independent expenditure presidential primary candidate state not saving

dheitzer:
[#1978](https://github.com/fecgov/fecfile-web-app/issues/1978)
  
The issue was that the candidate state was not saving on an independent expenditure with presidential primary.  This was by design as the candidate contact should not be updated for presidential primaries.  The solution was to add a candidate state field to the schedule e transaction and use that in this case.

## How to test

Add independent expenditure with candidate contact state set for a presidential primary.  Confirm this state appears when transaction is loaded.


mjtravers (commented):
@dheitzer When saving a senator or house representative, their state value is being saved in the transactions_schedulee.so_candidate_state field. If this is then changed in the contact record, the duplicate state records are out of sync. Would it be possible to add code to the ScheduleESerializer to make sure the so_candidate_state field is NULL if election_code does not start with "P" when doing a save?

mjtravers (approved):

Issue #816, opened 2024-04-10
Implement API Gateway architecture


lbeaufort:
Blocking #780

[...]

PR #913, opened 2024-06-13
Feature/816 - add reverse proxy

dheitzer:
Added nginx reverse proxy as part of the chosen approach for [ticket 816](https://github.com/fecgov/fecfile-web-api/issues/816)

This ticket also requires a network policy to be put in place as part of deployment.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -245,7 +245,7 @@ workflows:
             - dependency-check
           filters:
             branches:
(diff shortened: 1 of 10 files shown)


lbeaufort (commented):
Network policies should get added to the deploy task:

https://github.com/fecgov/fec-cms/blob/e95a584d372c9a2f800c3e981496e0df54ad17c3/tasks.py#L157-L170

mjtravers (approved):

Issue #918, opened 2024-06-18
Performance Improvements - Refactor report serializer can_delete() method


Elaine-Krauss-TCG:
The report serializer can_delete() method needs refactoring to improve performance: [https://github.com/fecgov/fecfile-web-api/blob/release/sprint-42/django-backend/fecfiler/reports/serializers.py#L144-L177](https://github.com/fecgov/fecfile-web-api/blob/release/sprint-42/django-backend/fecfiler/reports/serializers.py#L144-L177)

### QA Notes ###
Provide performance results in the comments.

Following the changes, there should be no visible changes when interacting with reports.  You should be able to load the Manage Reports table, load individual reports, and add and view Report Memos without issue.

[FECFILE-1430]([https://fecgov.atlassian.net/browse/FECFILE-1430|https://fecgov.atlassian.net/browse/FECFILE-1430])

### DEV Notes ###

The `to_representation` method in the report serializer has an out-sized impact on response times. The biggest factor here is the `can_delete` method, which can take upwards of 1,000ms to run when a committee has tens of thousands of transactions in it.

1. Improve performance of added annotation where used.
2. Remove addition of annotation when not needed in response.

### Design ###

null

[FECFILE-1430](https://fecgov.atlassian.net/browse/FECFILE-1430)

PR #950, opened 2024-07-03
feature/918 - Created can_delete column and trigger

sasha-dresden:
Issue #918


sasha-dresden:
Had to make some updates to the already existing django functionality to get it to activate the trigger on related reports.I also updated the trigger logic to be a bit easier to read and went back to how it was functioning in the django read version to update it in the trigger version

Elaine-Krauss-TCG (approved):

sasha-dresden:
Ok I think I cracked it! There were a few escapes I added to improve performance where it didn't make sense to run the update. The biggest of these was on transaction INSERT.
For the transaction INSERT 
 - Don't run if report is already marked as not able to delete. Adding a transaction would never flip it from not being able to delete to being able to delete.
 - Only run for the reports in the committee and year of the transaction.

However, the largest problem was our query we were using to find transactions in other reports. I had grabbed this query from what django had generated, but after some rework, converting to the subquery to a JOIN and using DISTINCT. This dramatically reduced the amount of time the query took to run. 

Both the original query and this query took approximately 50ms to run when there were less than 100 transactions, however, as the number of transactions grew, the original quickly grew. By 10k transactions it was taking over a 1 second, whereas the new one, while it does take a bit longer, grows much more slowly. By 10k it was only taking around 100ms. I didn't even attempt the first at 20k transactions, but the new one, was taking around 150ms, which seems reasonable. This was in a worst case scenario. Where all the reports where in the same committee, in the same year, and none of them had been marked as undeletable already.

mjtravers (approved):

PR #952, opened 2024-07-08
1458 - Fix itemization issue for independent expenditures

sasha-dresden:
Issue #953

### DIFF
--- a/django-backend/fecfiler/transactions/managers.py
+++ b/django-backend/fecfiler/transactions/managers.py
@@ -56,7 +56,19 @@ class TransactionManager(SoftDeleteManager):
             ),
             When(
                 transaction_type_identifier__in=schedule_e_over_two_hundred_types,
-                then=Q(_calendar_ytd_per_election_office__gt=Value(Decimal(200))),
+                then=Q(
+                    Q(
+                        parent_transaction__parent_transaction___calendar_ytd_per_election_office__gt=Value(  # noqa
+                            Decimal(200)
+                        )
+                    )
+                    | Q(
+                        parent_transaction___calendar_ytd_per_election_office__gt=Value(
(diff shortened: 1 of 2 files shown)


django-backend/fecfiler/transactions/managers.py
```diff
@@ -56,6 +54,10 @@ def ITEMIZATION_CLAUSE(self):  # noqa: N802
                 transaction_type_identifier__in=over_two_hundred_types,
                 then=Q(aggregate__gt=Value(Decimal(200))),
             ),
+            When(
+                transaction_type_identifier__in=schedule_e_over_two_hundred_types,
+                then=Q(_calendar_ytd_per_election_office__gt=Value(Decimal(200))),
+            ),
```
mjtravers, line 60: We need to look at the calendar_ytd_per_election_office and not the _calendar_ytd_per_election_office because we need to potentially look at the IE's parent transaction value for the IE memos.

Here is the code showing how the calendar_ytd_per_election_office is calculated:
https://github.com/fecgov/fecfile-web-api/blob/develop/django-backend/fecfiler/transactions/managers.py#L321-L325

Here is the code showing that a IE memo is part of the schedule_e_over_two_hundred_types:
https://github.com/fecgov/fecfile-web-api/blob/develop/django-backend/fecfiler/transactions/schedule_e/managers.py#L7-L9

mjtravers (changes requested):
Change needed when determining itemization based on > $200 value. See inline comments.

django-backend/fecfiler/transactions/managers.py
sasha-dresden: Updated this logic to check against all 3

mjtravers (approved):

PR #967, opened 2024-07-16
Feature/171 - Created performance testing framework to be able to run performance tests. Added performance test for creating a dot fec

sasha-dresden:
Issue [FECFILE-171](https://fecgov.atlassian.net/browse/FECFILE-171)
Created new TEST_RUNNER which silos tests marked with the performance tag. To use run:
python manage.py test --tag=performance

All tests marked @tag("performance") will be run (currently only the one).
By default python runs all tests, so I had to modify the test_runner so performance tagged tests wouldn't be run when no  tags were provided. 

Currently this can only be run locally, however a future ticket could look into conditionally running on other environments.

Findings:
I ran this test multiple times with an increasingly large number of transactions. Results below.
2k took 8.3 seconds
5k took 17.5 seconds
10k took 35.8 seconds
25k took 85.1 seconds

From these results it appears there might be some initial setup cost, but for the most part this is a linear process and higher numbers can be extrapolated. It seems to be able to process around 285 per second, however the greater the number of transactions the faster the rate, so I'm guessing it might be closer to 300 per second, so based off these estimates 1 million transactions would take a little under 1 hour. 

However, I didn't want to test further than 25k as the setup time was getting quite long. Creating 25k transactions took me 39 mins, however this was after I applied a performance improvement to the transaction aggregate trigger. Before this improvement 2k transactions took 20 minutes.
I did not include this trigger performance enhancement as part of this ticket as I did not want to have this PR be dependent upon that change.


lbeaufort (approved):
Looks great, thanks @sasha-dresden!

dheitzer (approved):
Maybe a future enhancement could be parameterizing the number of transactions to generate

PR #1005, opened 2024-08-05
feature/1431 updated docker compose memory limits to match cloud.gov

dheitzer:
https://fecgov.atlassian.net/browse/FECFILE-1431

Set for api, celery worker, and proxy to match cloud.gov manifests.


toddlees (approved):
Works on my local:
image

PR #1008, opened 2024-08-06
Feature/218

Elaine-Krauss-TCG:
[Fecfile #218](https://fecgov.atlassian.net/browse/FECFILE-218?atlOrigin=eyJpIjoiZmUyMGFhNjY3N2RkNDk4NjhiMTE4YmQzMTU1YmJkMzkiLCJwIjoiaiJ9)

### DIFF
--- a/django-backend/fecfiler/openfec/views.py
+++ b/django-backend/fecfiler/openfec/views.py
@@ -50,15 +50,17 @@ def retrieve_recent_f1(committee_id):
     First checks the realtime enpdpoint for a recent F1 filing.  If none is found,
     a request is made to a different endpoint that is updated nightly.
     The realtime endpoint will have more recent filings, but does not provide
-    filings older than 6 months. The nightly endpoint keeps a longer history"""
+    filings older than 6 months. The other endpoint returns a committee's current data
+    informed by their most recent F1 and F2 filings."""
     headers = {"Content-Type": "application/json"}
     params = {
         "api_key": settings.FEC_API_KEY,
         "committee_id": committee_id,
-        "sort": "-receipt_date",
-        "form_type": "F1",
     }
-    endpoints = [f"{settings.FEC_API}efile/filings/", f"{settings.FEC_API}filings/"]
+    endpoints = [
+        f"{settings.FEC_API}efile/form1/",
+        f"{settings.FEC_API}committee/{committee_id}/",
+    ]
     for endpoint in endpoints:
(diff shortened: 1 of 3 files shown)


django-backend/fecfiler/openfec/views.py
```diff
@@ -50,15 +50,17 @@ def retrieve_recent_f1(committee_id):
...
     headers = {"Content-Type": "application/json"}
     params = {
         "api_key": settings.FEC_API_KEY,
         "committee_id": committee_id,
-        "sort": "-receipt_date",
-        "form_type": "F1",
     }
-    endpoints = [f"{settings.FEC_API}efile/filings/", f"{settings.FEC_API}filings/"]
+    endpoints = [
+        f"{settings.FEC_API}efile/form1/",
```
toddlees, line 61: can you change this to ```f"{settings.FEC_API}efile/form1/?sort=-load_timestamp```
there can be multiple versions of the same f1, and we need the most recent.

toddlees (changes requested):

toddlees:
We're gonna need some follow up tickets.  This method `retrieve_recent_f1` is used in two places.
1) registering a committee account 
  - issue:  the f1 filing is used to get the email.  The second openfec endpoint `committee/{committee_id}` doesn't contain an email so this only works for that use case if the f1 filing is available in the first endpoint
2) creating the button that links to the f1 pdf on the committee info page
  - issue: here the f1 filing is used to get the pdf url.  again, the second endpoint doesn't have that info.  so it doesn't serve it's purpose if the filing is older than 6 months

django-backend/fecfiler/openfec/views.py
Elaine-Krauss-TCG: Sorry, I didn't see this.

If you look at the [https://api.open.fec.gov/developers/#/efiling/get_v1_efile_filings_](docs), the default sorting is `-receipt_date` which should satisfy that requirement.

Elaine-Krauss-TCG:
> We're gonna need some follow up tickets. This method `retrieve_recent_f1` is used in two places.
> 
>     1. registering a committee account
> 
>     * issue:  the f1 filing is used to get the email.  The second openfec endpoint `committee/{committee_id}` doesn't contain an email so this only works for that use case if the f1 filing is available in the first endpoint

The `committee/{committee_id}` [endpoint](https://api.open.fec.gov/developers/#/committee/get_v1_committee__committee_id__) does contain the email field, but it does _not_ contain the F1 filing url.

PR #1031, opened 2024-08-21
Feature/1541 - delete_committee_reports command

Elaine-Krauss-TCG:
[Fecfile#1541](https://fecgov.atlassian.net/browse/FECFILE-1541?atlOrigin=eyJpIjoiMjQ4NjM0MjY3N2RiNDY2MWJlMmNlYTZjNTYxMTQyM2QiLCJwIjoiaiJ9)


toddlees (approved):

django-backend/fecfiler/reports/management/commands/delete_committee_reports.py
```diff
@@ -0,0 +1,69 @@
...
+        self.stdout.write(self.style.SUCCESS(f"Dot Fecs: {dot_fec_count}"))
+        self.stdout.write(self.style.SUCCESS(
+            f"Upload Submissions: {upload_submission_count}"
+        ))
+        self.stdout.write(self.style.SUCCESS(
+            f"WebPrint Submissions: {web_print_submission_count}"
+        ))
+
+        reports.delete()
+        transactions.hard_delete()
```
mjtravers, line 69: @Elaine-Krauss-TCG  I think we can refactor this and the reports.views.e2e_delete_all_reports() method and reduce the code duplication.

mjtravers (changes requested):
See inline comments.

django-backend/fecfiler/reports/management/commands/delete_committee_reports.py
Elaine-Krauss-TCG: There are a few obstacles to using that code directly.  First, the function itself is inside the Reports Viewset.  It would be awkward to instantiate the Viewset just to use the e2e_delete_all_reports method, but it could be done.  The other obstacle is the logging.  Management commands use stdout to put messages in the console as opposed to the structlog logger.  There are ways around this, creating two separate wrapper methods that handle the logging in different ways for example, but they have their own inconveniences - passing the six different count values out.  None of these are difficult challenges, not at all, but at the time it seemed more inelegant than just having two functions.  I would be happy to revisit this if it would be helpful, though.

mjtravers (approved):

PR #1033, opened 2024-08-21
Feature/243 celery and db status

toddlees:
A viewset that provides actions to check the status of the system

The pattern here is that we check the cache for the status of some component.
If the cache is empty, we flag it with a `{}` and then process the status.
When the status has been determined, we update the cache with the result.
Whether the status comes from the cache or is determined on the fly, we return it.

The `{}` protects us from someone maliciously calling the endpoint to bog down the
system.  The side effect of that is that if you call the endpoint again before the
original request completes, you'll get the 503 response.  Calling the endpoint
multiple times rapidly is not a valid use case, so this is acceptable.

TODO: It would be ideal to have a scheduled task updating the cache rather than
the api request If we integrate something like Celery Beat, we could drastically
reduce the complexity of these endpoints to simply return the cache value.


dheitzer (approved):
looks nice.  I think the scheduled task is a good idea too.

mjtravers (approved):

PR #1052, opened 2024-08-28
feature/1552 - django key fallbacks

dheitzer:
## Summary

https://fecgov.atlassian.net/browse/FECFILE-1552

## How to test

New commands to support this functionality are documented in the devops app README with a SOP for both the django key and login cert rotation being uploaded to the shared drive. 

Unfortunately Django commands to not support sub directories so some devops commands were renamed for clarity.

Both commands tested in the develop environment.


Elaine-Krauss-TCG (approved):
Looks great :tada:

sasha-dresden:
This looks great but I think we might want some tests for the commands.  I'm not quite sure how challenging this would be to implement, but I did some quick research and it looks like we can use call_command and patch to simulate running these commands in a test
https://docs.djangoproject.com/en/5.1/ref/django-admin/#running-management-commands-from-your-code
https://docs.python.org/3/library/unittest.mock.html#unittest.mock.patch

dheitzer:
> This looks great but I think we might want some tests for the commands. I'm not quite sure how challenging this would be to implement, but I did some quick research and it looks like we can use call_command and patch to simulate running these commands in a test https://docs.djangoproject.com/en/5.1/ref/django-admin/#running-management-commands-from-your-code https://docs.python.org/3/library/unittest.mock.html#unittest.mock.patch

good catch - I added coverage to these cmds

PR #1084, opened 2024-09-12
Feature/1172 - Delete Transactions

sasha-dresden:
Issue [FECFILE-1172](https://fecgov.atlassian.net/browse/FECFILE-1172)


Elaine-Krauss-TCG (approved):

django-backend/fecfiler/transactions/views.py
```diff
@@ -149,10 +149,12 @@ def update(self, request, *args, **kwargs):
...
         return Response(saved_transaction.id)
 
-    def delete(self, request, *args, **kwargs):
-        response = super().delete(request, *args, **kwargs)
-        update_dependent_parent(self.get_object())
-        return response
+    def destroy(self, request, *args, **kwargs):
+        pk = kwargs.get("pk")
+        with db_transaction.atomic():
+            instance = Transaction.objects.get(pk=pk)
```
toddlees, line 155: In general we want to avoid using the base querysets in our views like `Transaction.objects.` and use the class's `get_queryset()` to preserve the access controls we have.   As this is written anyone could delete anyone's transaction, but if we use `get_queryset()` or `self.get_object()` (which depends on `get_queryset()`) we can limit access to the users' transactions.

We should have added tests for that so it would have been flagged after this change.  Maybe we can add some in.

django-backend/fecfiler/transactions/serializers.py
```diff
@@ -274,6 +274,7 @@ def to_representation(self, instance):
                 }
             )
 
+        representation["can_delete"] = instance.can_delete
```
toddlees, line 277: I think you can get rid of this line and add `can_delete = BooleanField(read_only=True)` to the fields above.  that will give us all the features of a serializer field

django-backend/fecfiler/transactions/models.py
```diff
@@ -179,6 +182,96 @@ def save(self, *args, **kwargs):
 
         super(Transaction, self).save(*args, **kwargs)
 
+    def delete(self):
+        if self.can_delete and not self.deleted:
+            self.deleted = datetime.now()
```
toddlees, line 187: because the model inherits SoftDeleteModel, you can replace 186,187 with `super().delete()`
Then all the other instances of 
```
x.deleted = datetime.now()
x.save()
```
can be replaced with `x.delete()`

toddlees (changes requested):

toddlees (approved):

PR #1094, opened 2024-09-20
Feature/1624 default committee to qualified

Elaine-Krauss-TCG:


Elaine-Krauss-TCG:
Not needed.  Change made on front-end instead

PR #1114, opened 2024-10-03
feature/1634 - josepy snyk report

dheitzer:
The approach chosen to remediating the snyk issue reported for josepy was to remove the josepy dependency altogether and replace it with jwcrypto.  This is beneficial because we just added the jwcrypto dependency to support our new mock_oidc_provider and also because jwcrypto is based on the cryptography package which is recommended over pyopenssl).  

This PR converts the josepy calls to the needed jwcrypto ones.  jwcrypto does more heavy lifting for us so some code could be removed as well.

https://fecgov.atlassian.net/browse/FECFILE-1634


sasha-dresden (approved):

Elaine-Krauss-TCG (approved):

lbeaufort (approved):
Excellent work @dheitzer!

PR #1121, opened 2024-10-09
feature/1680 - Fix issues with Schedule C and Schedule E not mapping to dot fec correctly

sasha-dresden:
Issue [FECFILE-1680](https://fecgov.atlassian.net/browse/FECFILE-1680)

So this can only be tested on stage which makes it a bit challenging to make sure it works. However, I believe I have all the issues resolved. I wasn't able to replicate the first issue on stage, so I think that one might've been fixed already, but for the others I believe it had to do with the dot fec composition and serialization. Some of these were showing an empty string when the value was NULL, when the it should display "0.00" so I needed to add some handling for that.
There were also a few items that weren't mapped correctly in the serializer and there was one date, for the loans, where the user can manually enter a date, so it's stored as a string instead of a date. I _think_ that's why it isn't displaying correctly there, so I tried to make sure it would get mapped in the same format as the other date objects. 
It's really hard to tell if this will work properly without actually being able to generate the pdf, so I focused on getting the dotfec to be what I think it should be for this to work.


django-backend/fecfiler/web_services/dot_fec/dot_fec_serializer.py
```diff
@@ -42,6 +43,30 @@ def date_serializer(model_instance, field_name, mapping):
...
+        model_instance, mapping.get("path", None) or field_name
+    )
+    for date_format in date_formats:
+        try:
+            # Try parsing with each format in the list
+            date_object = datetime.strptime(date_string, date_format).date()
+            return date_object.strftime("%Y%m%d") if date_object else ""
+        except ValueError:
+            continue  # If it fails, try the next format
+    return date_string
```
dheitzer, line 67: I feel like we might want to at least log an error here to alert us that a format was entered that we can't match on

dheitzer (changes requested):
wondering if we can log the failure scenario for the date format checking

dheitzer (approved):

dheitzer:
> wondering if we can log the failure scenario for the date format checking

addressed in a push I made

Elaine-Krauss-TCG (approved):

PR #1126, opened 2024-10-14
Revert "feature/1509 - create committee lookup"

Elaine-Krauss-TCG:
Reverts fecgov/fecfile-web-api#1123


Elaine-Krauss-TCG:
There was some confusion today around why committee data was being retrieved from EFO on DEV, STAGE, and Test.  I thought it might be code related, and that left this merge as the only realistic culprit, but ultimately that wasn't the case.

The real issue was that I loaded committee data on Friday _forgetting_ the all-important `--s3` flag...

PR #1149, opened 2024-10-25
Feature/1714 Itemize all Independent Expenditures

Elaine-Krauss-TCG:
[Fecfile#1714](https://fecgov.atlassian.net/browse/FECFILE-1714)


sasha-dresden (approved):
Looks good. One thing I might recommend changing if we're not going to use "over two hundred" going forward is changing the name of the over_two_hundred_types at line 47. Might be confusing when we come back to this in the future.

PR #1156, opened 2024-10-29
feature/1721 - Set reports.calculation_status to None whenever a transaction is updated

sasha-dresden:
Issue [FECFILE-1721](https://fecgov.atlassian.net/browse/FECFILE-1721)
APP [PR2311](https://github.com/fecgov/fecfile-web-app/pull/2311)

### DIFF
--- a/django-backend/fecfiler/transactions/views.py
+++ b/django-backend/fecfiler/transactions/views.py
@@ -305,6 +305,7 @@ class TransactionViewSet(CommitteeOwnedViewMixin, ModelViewSet):
                 schedule_instance.report_coverage_from_date = report.coverage_from_date
                 schedule_instance.save()
 
+        Report.objects.filter(id__in=report_ids).update(calculation_status=None)
         logger.info(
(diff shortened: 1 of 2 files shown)


dheitzer (approved):

django-backend/fecfiler/transactions/views.py
```diff
@@ -305,6 +305,7 @@ def save_transaction(self, transaction_data, request):
                 schedule_instance.report_coverage_from_date = report.coverage_from_date
                 schedule_instance.save()
 
+        Report.objects.all().update(calculation_status=None)
```
toddlees, line 308: Can you limit this to reports owned by the committee?

toddlees (changes requested):

django-backend/fecfiler/transactions/views.py
sasha-dresden: Good catch, got it fixed to only in committee

toddlees (approved):

PR #1159, opened 2024-10-31
feature/1715 - report status failure on submit polling

dheitzer:
https://fecgov.atlassian.net/browse/FECFILE-1715

DEV note on the ticket was correct in that this was caused only when polling for status happened.  We were getting the following fecfile error in the log which was causing it to get set to FAILURE:
`
fecfile-celery-worker    | [error    ] Error in polling: 'UploadSubmission' object has no attribute 'fec_batch_id' [fecfiler.web_services.tasks] _from_structlog=True _recor`

Solution was to add safe attribute access.  Also needed to update some function signatures to take fec_batch_id as they were inconsistent with the interface.  Added a unit test to test the function that was failing.


django-backend/fecfiler/web_services/tasks.py
```diff
@@ -229,16 +230,14 @@ def poll_for_fec_response(submission_id, submission_type_key, submission_name):
...
-        )
+        logger.info(f"Status: {submission.fec_status}, Message: {submission.fec_message}")
         logger.info(
             f"""Submission Polling - Attempt {
                 submission.fecfile_polling_attempts
             } / {EFO_POLLING_MAX_ATTEMPTS}"""
         )
         status_response_string = submitter.poll_status(
-            submission.fec_batch_id, submission.fec_submission_id
+            getattr(submission, "fec_batch_id", None), submission.fec_submission_id
```
toddlees, line 240: I think there's a better way to make poll_status generic.  Rather than having it take two different ids, one per type, lets have it take one id, and decide which to use in this method in the line before.

```
poll_id = getattr(submission, "fec_batch_id", None) or getattr(submission, "fec_submission_id", None)
status_response_string = submitter.poll_status(poll_id)
```

toddlees (changes requested):

django-backend/fecfiler/web_services/tasks.py
dheitzer: good catch.  @toddlees and I discussed and I updated the poll_status function to take a submission object to allow it to be generic as well as flexible with batch_id.

PR #1165, opened 2024-11-01
feature/1658 - Loans, debts, and repayments with future reports can't be deleted. Loans and debts with past reports can't be deleted.

sasha-dresden:
Issue [FECFILE-1658](https://fecgov.atlassian.net/browse/FECFILE-1658)
APP [PR2329](https://github.com/fecgov/fecfile-web-app/pull/2329)


django-backend/fecfiler/transactions/models.py
```diff
@@ -165,7 +165,24 @@ def children(self):
 
     @property
     def can_delete(self):
-        return len(self.blocking_reports) == 0
+        if len(self.blocking_reports) > 0:
+            return False
+        # For loan and debts, if there's a previous report you can't delete
+        if self.schedule_d or self.schedule_c:
```
dheitzer, line 171: I feel like we should use similar logic here to whatever we decide below for determining loan/debt so we don't have inconsistencies.

dheitzer (changes requested):

Elaine-Krauss-TCG:
The transactions cannot be deleted, as intended, but I am getting the "Delete" option for the Loan Made transaction.  In the API's response containing that transaction, "can_delete" is true, but actually trying to delete it fails because the parent can't be deleted.

PR #1189, opened 2024-11-08
feature/1558 - Add optional parameter to delete all contacts for committee

sasha-dresden:
Issue [FECFILE-1558](https://fecgov.atlassian.net/browse/FECFILE-1558)


Elaine-Krauss-TCG (approved):

dheitzer (commented):
Probably worth noting that this does a hard-delete (as opposed to the soft-delete functionality within the application) so we should not expect to be able to restore any of these removed contacts.

dheitzer (approved):

toddlees (approved):

PR #1204, opened 2024-11-14
feature/1752 - Update committee and candidate validation to limit to alphanumeric values and then use the sanitized candidate/committee id for building the url

sasha-dresden:
Issue [FECFILE-1752](https://fecgov.atlassian.net/browse/FECFILE-1752)

### DIFF
--- a/django-backend/fecfiler/contacts/views.py
+++ b/django-backend/fecfiler/contacts/views.py
@@ -40,7 +40,19 @@ FEC_API_CANDIDATE_ENDPOINT = (
 def validate_and_sanitize_candidate(candidate_id):
     if candidate_id is None:
         raise AssertionError("No Candidate ID provided")
-    return candidate_id
+    if re.match(r"^[a-zA-Z0-9]+$", candidate_id):
+        return candidate_id
+    else:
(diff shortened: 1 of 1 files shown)


django-backend/fecfiler/contacts/views.py
```diff
@@ -40,7 +40,19 @@
...
+    if re.match(r"^[a-zA-Z0-9]+$", candidate_id):
+        return candidate_id
+    else:
+        raise AssertionError("Candidate ID provided invalid")
+
+
+def validate_and_sanitize_committee(committee_id):
+    if committee_id is None:
+        raise AssertionError("No Committee ID provided")
+    if re.match(r"^[a-zA-Z0-9]+$", committee_id):
```
toddlees, line 52: can you update both of these to also be 9 characters

toddlees (changes requested):

django-backend/fecfiler/contacts/views.py
sasha-dresden: Ok got it updated so it'll only accept if user puts in exactly 9 characters for those.

PR #1211, opened 2024-11-16
1765 - Add temporary default value to non-nullable fields in `CashOnHandYearly` to fix migration

lbeaufort:
Ticket link: https://fecgov.atlassian.net/jira/software/c/projects/FECFILE/issues/FECFILE-1765
  
Related PRs: None

`preserve_default=False` in migration

https://docs.djangoproject.com/en/5.1/ref/migration-operations/#alterfield
>The preserve_default argument indicates whether the field’s default value is permanent and should be baked into the project state (True), or if it is temporary and just for this migration (False) - usually because the migration is altering a nullable field to a non-nullable one and needs a default value to put into existing rows

Verified this won't impact dev/stage/prod because none of the fields are null:
`select count(*) from cash_on_hand_yearly where cash_on_hand is null or year is null;`


dheitzer (approved):

django-backend/fecfiler/cash_on_hand/migrations/0002_alter_cashonhandyearly_cash_on_hand_and_more.py
```diff
@@ -0,0 +1,25 @@
...
+        migrations.AlterField(
+            model_name='cashonhandyearly',
+            name='cash_on_hand',
+            field=models.DecimalField(decimal_places=2, default=0, max_digits=11),
+            preserve_default=False,
+        ),
+        migrations.AlterField(
+            model_name='cashonhandyearly',
+            name='year',
+            field=models.TextField(default=2020, unique=True),
```
toddlees, line 22: I think we need to remove the unique as well.  Two committees can't override the same year otherwise.

toddlees (changes requested):

django-backend/fecfiler/cash_on_hand/migrations/0002_alter_cashonhandyearly_cash_on_hand_and_more.py
lbeaufort: Thanks @toddlees! Change made in https://github.com/fecgov/fecfile-web-api/pull/1211/commits/98babe80ce3ab904785c49f99def1ffc088756d0 and test added in https://github.com/fecgov/fecfile-web-api/pull/1211/commits/079c0d4515db625ae6a1dcc9256f8024e1a4c92f. Please let me know if you want me to make any other changes.

toddlees: I think you can get rid of the unique in 0002 and no longer need 0003 as it's own migration

toddlees (changes requested):

django-backend/fecfiler/cash_on_hand/migrations/0002_alter_cashonhandyearly_cash_on_hand_and_more.py
lbeaufort: Change made in https://github.com/fecgov/fecfile-web-api/pull/1211/commits/12392df95bf84003aad4bccb47d0c8c79f903351

toddlees (approved):

PR #1234, opened 2024-12-03
Move deploy notes to central wiki

toddlees:


lbeaufort:
@toddlees this is a nice change! One small thought - if we named the wiki `Deployment` it would have a much friendlier URL.

lbeaufort (approved):
Thanks @toddlees!

PR #1241, opened 2024-12-06
Feature/513 itemization updates

dheitzer:
Ticket link: https://fecgov.atlassian.net/browse/FECFILE-513
  
Related PRs:


dheitzer:
@toddlees @Elaine-Krauss-TCG @sasha-dresden this one is ready for CR

sasha-dresden (dismissed):
This looks really good! There's only two things I noticed. The first is exceptionally minor. the name of your trigger is: zafter_transactions_transaction_insert_or_update_trigger
And I'm guessing that leading "z" wasn't supposed to be there.

The second thing is for your trigger there's no reverse function. I know I forgot about this almost always, but after our meeting last week about migration reversibility I think it might be a good idea to make sure all of our migrations have reverse in them. A good example of one that I remembered to do it in is here: 
https://github.com/fecgov/fecfile-web-api/blob/85a79c4cc5d408ab3861e37a5c9134b416baacc6/django-backend/fecfiler/reports/migrations/0009_report_can_delete.py#L178

This would would work well for the populate_over_two_hundred_types function to reverse it but for your triggers their inside a RunSQL. I think if you use RunSQL you would need to do something like this instead:
 migrations.RunSQL(
            """
            Create Triggers and functions sql
            """,
            reverse_sql="""
            Drop Triggers and functions sql
            """

dheitzer:
removed the ` CREATE EXTENSION IF NOT EXISTS "uuid-ossp";` as pointed out by @toddlees

dheitzer:
> This looks really good! There's only two things I noticed. The first is exceptionally minor. the name of your trigger is: zafter_transactions_transaction_insert_or_update_trigger And I'm guessing that leading "z" wasn't supposed to be there.
> 
> The second thing is for your trigger there's no reverse function. I know I forgot about this almost always, but after our meeting last week about migration reversibility I think it might be a good idea to make sure all of our migrations have reverse in them. A good example of one that I remembered to do it in is here:
> 
> https://github.com/fecgov/fecfile-web-api/blob/85a79c4cc5d408ab3861e37a5c9134b416baacc6/django-backend/fecfiler/reports/migrations/0009_report_can_delete.py#L178
> 
> This would would work well for the populate_over_two_hundred_types function to reverse it but for your triggers their inside a RunSQL. I think if you use RunSQL you would need to do something like this instead: migrations.RunSQL( """ Create Triggers and functions sql """, reverse_sql=""" Drop Triggers and functions sql """

@sasha-dresden yes the prepended 'z' was intentional as discussed in the dev meeting to ensure this trigger gets executed after calculate_aggregates.  I also added reversible migrations per your suggestion and example - ty!

toddlees (approved):

PR #1249, opened 2024-12-10
feature/1606 - Add ENABLE_DEVELOPER_COMMANDS environment variable and limit manage.py subcommands

sasha-dresden:
Issue [FECFILE-1606](https://fecgov.atlassian.net/browse/FECFILE-1606)


django-backend/fecfiler/settings/base.py
```diff
@@ -380,3 +380,5 @@ def get_logging_processors():
 )
 
 TEST_RUNNER = "fecfiler.test_runner.CustomTestRunner"
+
+ENABLE_DEVELOPER_COMMANDS = env.get_credential("ENABLE_DEVELOPER_COMMANDS")
```
toddlees, line 384: Can you make the right side evaluate to a boolean?  Right now it's pulling a string which is truthy unless "" or None

toddlees (changes requested):

django-backend/fecfiler/settings/base.py
sasha-dresden: Oh good catch! Yep got that fixed now.

Elaine-Krauss-TCG (approved):

toddlees (approved):

PR #1264, opened 2024-12-18
Feature/1824 - temporary sec consent

dheitzer:
Ticket link: https://fecgov.atlassian.net/browse/FECFILE-1824
  
Related PRs: https://github.com/fecgov/fecfile-web-app/pull/2456

### DIFF
--- a/django-backend/fecfiler/user/views.py
+++ b/django-backend/fecfiler/user/views.py
@@ -1,38 +1,31 @@
 from rest_framework.response import Response
 from rest_framework.viewsets import ModelViewSet
-import datetime
+from .serializers import CurrentUserSerializer, session_security_consented_key
+from datetime import date, timedelta
 import logging
 
 logger = logging.getLogger(__name__)
-
-session_security_consented_key = "session_security_consented"
 
 
 class UserViewSet(ModelViewSet):
 
     def retrieve(self, request, *args, **kwargs):
         serializer = self.get_serializer(request.user)
-        augmented_response = {
-            "security_consented": (
-                datetime.datetime.now() <= request.user.security_consent_exp_date
-                or request.session[session_security_consented_key]
-            )
-        }
-        augmented_response.update(serializer.data)
-        return Response(augmented_response)
+        return Response(serializer.data)
 
     def update(self, request, *args, **kwargs):
-        add_session_consent = False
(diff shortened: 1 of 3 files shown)


django-backend/fecfiler/user/views.py
```diff
@@ -1,22 +1,31 @@
...
-        if request.method == 'GET':
-            serializer = self.get_serializer(request.user)
-            return Response(serializer.data)
-        if request.method == 'PUT':
-            serializer = self.get_serializer(request.user, data=request.data)
-            if serializer.is_valid():
-                serializer.save()
-                return Response(serializer.data)
-            return HttpResponseBadRequest()
+class UserViewSet(ModelViewSet):
```
toddlees, line 10: I think ModelViewSet has more mixins than we want for the user view.  The old had the scope narrowed to CRU without Delete or List.  You can add them in manually
```
class ModelViewSet(mixins.CreateModelMixin,
                   mixins.RetrieveModelMixin,
                   mixins.UpdateModelMixin,
                   mixins.DestroyModelMixin,
                   mixins.ListModelMixin,
                   GenericViewSet):
```

django-backend/fecfiler/user/views.py
```diff
@@ -1,22 +1,31 @@
...
+
+    def retrieve(self, request, *args, **kwargs):
+        serializer = self.get_serializer(request.user)
+        return Response(serializer.data)
+
+    def update(self, request, *args, **kwargs):
```
toddlees, line 16: I think the front end is still expecting to work with `current`
image

toddlees (changes requested):

django-backend/fecfiler/user/views.py
dheitzer: good catch! removed ModelViewSet and used just GenericViewSet with the use of actions

django-backend/fecfiler/user/views.py
dheitzer: good catch, not sure how this slipped past my testing.  Updated the urls to get_current and update_current to avoid needing to set a user id in the url.

Elaine-Krauss-TCG (approved):

toddlees (approved):

PR #1270, opened 2025-01-07
Release/sprint 53

toddlees:


lbeaufort:
Do we care about https://github.com/fecgov/fecfile-web-api/pull/1270/checks?check_run_id=35256545924 duplicated code warnings on tests? Arguably it's not that bad to duplicate code on tests so could consider exempting thme, but also we could potentially make them more maintainable/efficient if we addressed it.

lbeaufort (approved):
Approved, with one question: https://github.com/fecgov/fecfile-web-api/pull/1270#issuecomment-2590185529

PR #1280, opened 2025-01-14
Feature/1869 - Remove committee transaction views from DB

sasha-dresden:
Issue [FECFILE-1869](https://fecgov.atlassian.net/browse/FECFILE-1869)


django-backend/fecfiler/committee_accounts/README.md
```diff
@@ -4,31 +4,6 @@ When changes are made to the Committee Account model, it is necessary to update
 the Committee Views for each committee in the database. To do this, create a new
```
dheitzer, line 4: Maybe this whole "Updating Committee Views in the Database" section can be removed

dheitzer (commented):
This looks great.  The only question I would have is whether we would want to change the `transaction_view` manager function to a new name to avoid future confusion about the use of views.  Perhaps like annotated_transaction?

django-backend/fecfiler/transactions/managers.py
```diff
@@ -231,6 +230,12 @@ def PAYMENT_AMOUNT_CLAUSE(self):  # noqa: N802
...
     def transaction_view(self):
+        REPORT_CODE_LABEL_CLAUSE = Subquery(  # noqa: N806
+            Report.objects.filter(transactions=OuterRef("pk"))
+            .annotate(report_code_label=report_code_label_case)
+            .values("report_code_label")[:1]
+        )
+
         return (
             super()
             .get_queryset()
```
dheitzer, line 241: Do we need effective_amount to be here anymore? seems like it's only used within the triggers functions which redfine it themselves now.

dheitzer (changes requested):
found a potential couple cleanup items

Elaine-Krauss-TCG (approved):

django-backend/fecfiler/committee_accounts/README.md
sasha-dresden: Good catch, we definitely don't need this if committee views don't exist. Removing it.

django-backend/fecfiler/transactions/managers.py
sasha-dresden: Another good catch. We don't need this and it also slows it down a bit when we do use the transaction_view(). Removing both of them.

sasha-dresden:
Oh, @toddlees I know we spoke a bit about potentially merging the transaction_view() into the regular get_queryset(). I didn't do any extensive testing on it, but when I tried to do so my tests went from taking 80 - 85 seconds to taking in the 130 second range. Maybe we could optimize this queryset eventually, but merging as is definitely seems to slow it down.

PR #1288, opened 2025-01-16
feature/1940 - Add celery beat and a test task

sasha-dresden:
Issue [FECFILE-1940](https://fecgov.atlassian.net/browse/FECFILE-1940)

### DIFF
--- a/django-backend/fecfiler/web_services/summary/tasks.py
+++ b/django-backend/fecfiler/web_services/summary/tasks.py
@@ -1,5 +1,6 @@
 from enum import Enum
 from celery import shared_task
+from django.db import connection
 from fecfiler.reports.models import Report, FORMS_TO_CALCULATE
 from .summary import SummaryService
 import uuid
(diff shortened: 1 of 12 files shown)


lbeaufort (commented):
Thanks @sasha-dresden! Initial review looks good - did you test with a manual deploy to make sure everything works in cloud.gov?

django-backend/fecfiler/web_services/summary/tasks.py
```diff
@@ -267,3 +268,39 @@ def calculate_summary(report_id):
             logger.info(f"Report: {report.id} recalculated")
 
     return primary_report.id
+
+
+@shared_task
+def get_database_connections():
```
toddlees, line 274: lets move this out of the summary folder and into a new devops folder

toddlees (changes requested):
Let's pick between 'celery-beat' and 'scheduler'.
I think if we change the name of the run script to run-scheduler.sh and update the docker service name to scheduler, we should be using that name consistently everywhere

django-backend/fecfiler/web_services/summary/tasks.py
sasha-dresden: Oh I thought that was a wierd place to put it. I was looking at the previous attempt and that's where it was. devops makes much more sense. We already had a devops folder, but no tasks file so I created one and moved it there.

sasha-dresden:
> Thanks @sasha-dresden! Initial review looks good - did you test with a manual deploy to make sure everything works in cloud.gov?

Just did. Needed to make a few changes to the manifest and I needed to make the script file executable. Apparently it was fine as it was running it through docker but needed updated permissions for running it in cloud. But was able to have a successful deploy to dev with it running and kicking off its task to the web-services celery.

toddlees (approved):

PR #1323, opened 2025-02-10
FECFILE-1992: disable_user custom command

danguyf:
Ticket link:
  https://fecgov.atlassian.net/browse/FECFILE-1992

Related PRs:
None.

I deliberated a considerable while on what the command should be -- enable_user, disable_user, toggle_user, user_set_active with an argument for active/inactive, etc.  Eventually I decided that this was simplest: the primary use will be disabling the user, which it defaults to, and a flag can be passed in to re-enable.


django-backend/fecfiler/user/management/commands/disable_user.py
```diff
@@ -0,0 +1,28 @@
...
+
+
+class Command(BaseCommand):
+    help = 'Disable (or re-enable) a user with a given email.'
+
+    def add_arguments(self, parser):
+        parser.add_argument('email', type=str,
+                help='The email address of the user to be disabled/enabled.')
+        # add an --enable flag that defaults to false if not found
+        parser.add_argument('-d', '--enable', action='store_true',
```
dheitzer, line 12: I wonder if we should change -d to -e ? just for consistency.

dheitzer (dismissed):
just a couple comments.  otherwise looks good!

django-backend/fecfiler/committee_accounts/utils.py
```diff
@@ -94,9 +69,6 @@ def check_can_create_committee_account(committee_id, user):
     if existing_account:
         failure_reason = f"Committee account {committee_id} already created"
 
-    if not check_email_allowed_to_create_committee_account(email):
```
dheitzer, line 97: was this and associated function/test supposed to be removed?

django-backend/fecfiler/committee_accounts/utils.py
```diff
@@ -34,31 +34,6 @@ def get_response_for_bad_committee_source_config():
...
-    or exception list.
-
-    Args:
-        email (str): The email to be checked.
-
-    Returns:
-        boolean True if email is allowed, False otherwise.
-    """
-    allowed_domains = ["fec.gov"]
-    allowed_emails = settings.CREATE_COMMITTEE_ACCOUNT_ALLOWED_EMAIL_LIST
```
dheitzer, line 49: we can probably remove this CREATE_COMMITTEE_ACCOUNT_ALLOWED_EMAIL_LIST settings/env var as well.

django-backend/fecfiler/committee_accounts/utils.py
```diff
@@ -34,31 +34,6 @@ def get_response_for_bad_committee_source_config():
     return response
 
 
-def check_email_allowed_to_create_committee_account(email):
```
dheitzer, line 37: I think we still want the retain the functionality for requiring a fec.gov email in this function, correct?.  I think only functionality we're removing would be the check on line 49.

django-backend/fecfiler/user/management/commands/disable_user.py
danguyf: Oops!  That was an artifact from when the argument was --disable, not --enable, good catch.  Done.

django-backend/fecfiler/committee_accounts/utils.py
danguyf: Makes sense. Done.

django-backend/fecfiler/committee_accounts/utils.py
lbeaufort: @dheitzer great question! We do not want to keep the requirement that users have an FEC email address.

django-backend/manage.py
```diff
@@ -31,6 +31,7 @@
     "delete_committee_reports",
     # users
     "reset_security_consent_date",
+    "disable_user"
```
lbeaufort, line 34: @danguyf We'll need to be able to do this in production, so I think it needs to be a `core` command (see line 9)

cc: @toddlees in case you disagree

toddlees: yes core is the correct list for this to go in

danguyf: Done!

toddlees (approved):

toddlees:
Great solution!  exit code 1 on errors is perfect too because we can see a task as failed if we run `cf run-task "python manage.py disable_user xxx`.

PR #1329, opened 2025-02-13
Updates locust_run to be compatible with remote testing again

Elaine-Krauss-TCG:
Ticket link: N/A

### DIFF
--- a/performance-testing/locust_run.py
+++ b/performance-testing/locust_run.py
@@ -1,5 +1,4 @@
 import os
-import resource
 import logging
 import random
 import json
@@ -26,9 +25,6 @@ SINGLE_TO_TRIPLE_RATIO = float(
 )
 
 SCHEDULES = ["A"]  # Further schedules to be implemented in the future
-
-# Avoid "Too many open files" error
-# resource.setrlimit(resource.RLIMIT_NOFILE, (10000, 999999))
 
 
 def get_json_data(name):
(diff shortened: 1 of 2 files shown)


performance-testing/locust_run.py
```diff
@@ -27,7 +28,7 @@
 SCHEDULES = ["A"]  # Further schedules to be implemented in the future
 
 # Avoid "Too many open files" error
-resource.setrlimit(resource.RLIMIT_NOFILE, (10000, 999999))
+# resource.setrlimit(resource.RLIMIT_NOFILE, (10000, 999999))
```
dheitzer, line 31: maybe we can remove this if no longer needed

dheitzer (changes requested):
just one minor request, otherwise looks great! nice job

performance-testing/locust_run.py
Elaine-Krauss-TCG: This line wouldn't work in my environment, but it also wasn't needed for locust to work.  I didn't want to remove it without knowing what it was intended to do, though

lbeaufort: @Elaine-Krauss-TCG thanks for taking a look. This is legacy from openFEC - let's leave it commented out as a breadcrumb in case someone else has a different issue.

lbeaufort (approved):

performance-testing/locust_run.py
lbeaufort: After further discussion, we agreed it's fine to take it out for now.

PR #1369, opened 2025-03-05
1706 Production Committee Data Flow

toddlees:
Ticket link: https://fecgov.atlassian.net/browse/FECFILE-1706


django-backend/fecfiler/committee_accounts/utils.py
```diff
@@ -90,37 +76,129 @@ def create_committee_account(committee_id, user):
...
+def get_raw_committee_data(committee_id):
+    params = {
+        "api_key": settings.PRODUCTION_OPEN_FEC_API_KEY,
+        "committee_id": committee_id,
+    }
+    committee_data = query_fec_api(
+        f"{settings.PRODUCTION_OPEN_FEC_API}efile/form1/", params
     )
 
+    if committee_data:
```
dheitzer, line 185: is it worth factoring out this section into a shared function to be used here and in get_committee_from_test_fec() ?

dheitzer (approved):
looks great! just one (potential) enhancement

django-backend/fecfiler/committee_accounts/utils.py
toddlees: good point i was thinking the same thing, but was betting that there would be differences that would force us to split them up again. After speaking with Sasha this afternoon, it looks like there will be differences in this snippet from the test one.

PR #1382, opened 2025-03-10
FECFILE-2070: Added option to disable by UUID.

danguyf:
Ticket link: https://fecgov.atlassian.net/browse/FECFILE-2070

Related PRs:

Previously, the disable_user command required that an email be passed in. Now you may pass in an email OR a UUID, though you must pass in one.

### DIFF
--- a/django-backend/fecfiler/user/management/commands/disable_user.py
+++ b/django-backend/fecfiler/user/management/commands/disable_user.py
@@ -32,6 +32,6 @@ class Command(BaseCommand):
         user.save()
 
         self.stdout.write(self.style.SUCCESS(
-            f'The is_active flag for user [{user.username} | {user.email}] ' +
+            f'The is_active flag for user [{user.username} | {user.email}] '
             f'set to: {user.is_active}'
         ))
--- a/django-backend/fecfiler/user/management/commands/disable_user.py
+++ b/django-backend/fecfiler/user/management/commands/disable_user.py
@@ -6,7 +6,11 @@ class Command(BaseCommand):
     help = 'Disable (or re-enable) a user with a given email.'
 
     def add_arguments(self, parser):
-        parser.add_argument('email', type=str,
+        # require either a UUID or email
+        id_arg = parser.add_mutually_exclusive_group(required=True)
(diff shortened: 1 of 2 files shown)


django-backend/fecfiler/user/management/commands/disable_user.py
```diff
@@ -16,13 +20,18 @@ def handle(self, *args, **options):
         user_model = get_user_model()
 
         try:
-            user = user_model.objects.get(email=options['email'])
+            # if they use both arguments, prefer UUID
+            if options['uuid'] is not None:
+                user = user_model.objects.get(username=options['uuid'])
```
toddlees, line 25: This might work locally if the uuid in the username is the same as the pk id for the user record.  In our deployments, the uuid in username holds the login.gov uuid rather the the uuid which is the pk/id of the user record in our database.  We want the later for this feature.

django-backend/fecfiler/user/management/commands/disable_user.py
```diff
@@ -16,13 +20,18 @@ def handle(self, *args, **options):
...
+                user = user_model.objects.get(email=options['email'])
         except user_model.DoesNotExist:
             raise CommandError('User does not exist')
 
         user.is_active = options['enable']
         user.save()
 
         self.stdout.write(self.style.SUCCESS(
-            f'The is_active flag for user [{user.email}] set to: {user.is_active}'
+            f'The is_active flag for user [{user.username} | {user.email}] '
```
toddlees, line 35: we'll want this to use the user.id field as well

toddlees (changes requested):

django-backend/fecfiler/user/management/commands/disable_user.py
danguyf: Switched from user.username (login.gov UUID) to user.id (FECFile+ UUID).

django-backend/fecfiler/user/management/commands/disable_user.py
danguyf: Switched from user.username (login.gov UUID) to user.id (FECFile+ UUID).

toddlees (approved):

PR #1383, opened 2025-03-12
[Hold for Todd] Feature/2087 aggregation fix

Elaine-Krauss-TCG:
Ticket link: [here](https://fecgov.atlassian.net/browse/FECFILE-2087)
  
Related PRs: [App PR](https://github.com/fecgov/fecfile-web-app/pull/2595)

## Note: Ryan Lanz wants to be notified when this gets merged to DEV


Elaine-Krauss-TCG:
Note: Ryan Lanz wants to be notified when this gets merged to DEV

danguyf (commented):
I understand it all but transactions/views.py:387-439, I'm still piecing together my understanding of that block.  I'd approve it otherwise!

danguyf (approved):
I grok it all now.  No questions, suggestions, or concerns.

django-backend/fecfiler/transactions/views.py
```diff
@@ -234,16 +235,26 @@ def get_previous(
...
             query = query.filter(
-                Q(created__lt=transaction.created) | ~Q(date=date)
+                Q(created__lt=original_transaction.created) | ~Q(date=date)
             )
 
         query = query.order_by("-date", "-created")
         previous_transaction = query.first()
 
         if previous_transaction:
+            if original_transaction and (
```
toddlees, line 250: I don't think we need to be changing anything for this to work.  The only data that matters from the existing transaction is its created date.  The rest is entirely derived from the query params

toddlees (changes requested):

toddlees (changes requested):

django-backend/fecfiler/transactions/views.py
```diff
@@ -356,6 +384,59 @@ def save_transaction(self, transaction_data, request):
         update_dependent_children(transaction_instance)
         delete_carried_forward_loans_if_needed(transaction_instance, committee_id)
         delete_carried_forward_debts_if_needed(transaction_instance, committee_id)
+
+        # Set the earlier date in order to detect when a transaction has moved forward
+        if original_date and original_date < schedule_instance.get_date():
+            transactions_after_original = Transaction.objects.get_queryset().filter(
```
toddlees, line 390: I like the idea of using django's queryset to build a common result set.  The issue i see here is that you'll need to use a different set than how you've defined `transactions_after_original` as a base for finding the next transaction by election, since the next transaction by election doesn't care who contact_1 is.

I might sugest something like:
```
transactions_after_original = (queryset..).filter(same year, after-or-created-after)
next_transaction_by_entity = transactions_after_original.filter(contact_1=original_contact_1)[:0]
if schedule e:
  next_transaction_by_election = transactions_after_original.filter(election matching stuff)[:0]
...
```

The nice thing about election aggregate is that it only needs to happen over schedule e transactions

Elaine-Krauss-TCG: The set of election transactions is no longer a subset of `transactions_after_original` and I've cut it down to just schedule e transactions

sasha-dresden (approved):
There is a *lot* going on with these, always is with aggregates. But nothing jumps out at me as incorrect or anything and the tests seem incredibly thorough and covering every case so I think it should be good. My one concern is with the querysets, I've been bitten a few times with performance woes due to not being aware of hidden costs in the querysets so I'd be interested in seeing how expensive those transactions_after_original and election_entities querysets are. I do really like how they are contained within if statements so they only run when they absolutely have to. Otherwise this looks amazing!

danguyf (approved):

toddlees (approved):

toddlees:
> There is a _lot_ going on with these, always is with aggregates. But nothing jumps out at me as incorrect or anything and the tests seem incredibly thorough and covering every case so I think it should be good. My one concern is with the querysets, I've been bitten a few times with performance woes due to not being aware of hidden costs in the querysets so I'd be interested in seeing how expensive those transactions_after_original and election_entities querysets are. I do really like how they are contained within if statements so they only run when they absolutely have to. Otherwise this looks amazing!

Agreed!  made ticket to investigate with silk. https://fecgov.atlassian.net/browse/FECFILE-2137

PR #1405, opened 2025-03-26
Add cert files to gitignore

lbeaufort:
Ticket link: None
 
When we generate new certificate files for login.gov, we should make sure they aren't accidentally checked in to source control.


dheitzer (approved):
great idea

PR #1407, opened 2025-03-27
feature/674 - Allow user to create Coordinated Party Expenditures on Schedule F

dheitzer:
Ticket link: https://fecgov.atlassian.net/browse/FECFILE-674
  
Related PRs: https://github.com/fecgov/fecfile-validate/pull/342 https://github.com/fecgov/fecfile-web-app/pull/2620


django-backend/fecfiler/transactions/schedule_f/utils.py
```diff
@@ -10,6 +10,7 @@ def add_schedule_f_contact_fields(instance: Transaction, representation=None):
         add_candidate_contact(data, instance.contact_2, "payee", False)
     if instance.contact_3:
         data["payee_committee_id_number"] = instance.contact_3.committee_id
+        data["payee_committee_name"] = instance.contact_3.committee_id
```
Elaine-Krauss-TCG, line 13: I feel like this might be a typo?

Elaine-Krauss-TCG: Follow-up: loading a saved Schedule F transaction shows that the committee id is being saved in the committee name field
image

sasha-dresden:
Is there a reason for  linter.IgnoreMigration()?

django-backend/fecfiler/transactions/schedule_f/utils.py
dheitzer: great catch! @Elaine-Krauss-TCG  #eagle-eye.  Thanks for testing things.  Pushed this fix up.

dheitzer:
> Is there a reason for linter.IgnoreMigration()?

@sasha-dresden it was complaining that the migration was not reversible.  Since schF is new functionality however, I don't believe it's necessary.  We to not add the feature flag to the schF api work even if feature flag is disabled in the app too, so I can't think of a reason to we would ever need to reverse the migration.

sasha-dresden:
> > Is there a reason for linter.IgnoreMigration()?
> 
> @sasha-dresden it was complaining that the migration was not reversible. Since schF is new functionality however, I don't believe it's necessary. Even the api work is not behind a feature flag per design decision, so I can't think of a reason we would ever need to reverse the migration.

Ah that makes sense. Yes I had quite a fight with the linter when making the memo_text field non-nullable. But that one was not behind a feature flag. So I think you're right that it's ok here.

sasha-dresden (approved):

PR #1410, opened 2025-03-31
Feature/2071 remove email logging

Elaine-Krauss-TCG:
Ticket link:
  
Related PRs:

### DIFF
--- a/django-backend/fecfiler/committee_accounts/views.py
+++ b/django-backend/fecfiler/committee_accounts/views.py
@@ -212,8 +212,8 @@ class CommitteeMembershipViewSet(CommitteeOwnedViewMixin, viewsets.ModelViewSet)
         new_member = Membership(**membership_args)
         new_member.save()
 
-        member_type = "existing user" if user else "pending membership for"
-        logger.info(f'Added {member_type} "{email}" to committee {committee} as {role}')
+        member_type = f'existing user {user.id} to' if user else "pending membership for"
+        logger.info(f'User {request.user.id} added {member_type} committee {committee} as {role}')
 
         return Response(CommitteeMembershipSerializer(new_member).data, status=200)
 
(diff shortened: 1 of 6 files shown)


danguyf (approved):

lbeaufort (dismissed):
Re: disable user logging

django-backend/fecfiler/committee_accounts/views.py
```diff
@@ -239,7 +241,7 @@ def remove_member(self, request, pk: UUID):
         try:
             member.delete()
             logger.info(
-                f"{request.user.email} removed {member.email} "
+                f"{request.user.id} removed {member.id} "
```
toddlees, line 244: I think `member.id` needs to be `member.user.id`.  

Just a thought, not necessary but, might be worth considering renaming the local variable `member` to `membership` if it's getting confused with representing a user

django-backend/fecfiler/committee_accounts/views.py
```diff
@@ -263,11 +265,13 @@ def update(self, request, *args, **kwargs):
             )
 
         existing_member = self.get_object()
-        email = existing_member.email
+        user_id = existing_member.id
```
toddlees, line 268: same here

django-backend/fecfiler/user/management/commands/disable_user.py
```diff
@@ -32,6 +32,5 @@ def handle(self, *args, **options):
         user.save()
 
         self.stdout.write(self.style.SUCCESS(
-            f'The is_active flag for user [{user.id} | {user.email}] '
-            f'set to: {user.is_active}'
+            f'The is_active flag for user {user.id} has been set to: {user.is_active}'
```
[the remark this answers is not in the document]
toddlees: agreed that we can make the exception here.  We're essentially opting the info out of obscurity because it's for an incident.

toddlees (changes requested):

django-backend/fecfiler/committee_accounts/views.py
Elaine-Krauss-TCG: Oh, that's a great catch!  On it~

toddlees (approved):

PR #1416, opened 2025-04-02
Feature/form 3

toddlees:
Ticket link: https://fecgov.atlassian.net/browse/FECFILE-1979
VAL:https://github.com/fecgov/fecfile-validate/pull/346
APP:https://github.com/fecgov/fecfile-web-app/pull/2638
API:https://github.com/fecgov/fecfile-web-api/pull/1416


django-backend/fecfiler/reports/form_3/serializers.py
```diff
@@ -0,0 +1,299 @@
...
+    ValidationError,
+    DecimalField,
+    DateField,
+    BooleanField,
+)
+import structlog
+
+logger = structlog.get_logger(__name__)
+
+COVERAGE_DATE_REPORT_CODE_COLLISION = ValidationError(
```
sasha-dresden, line 18: I don't know if we're going to be using this validation error with other forms, but we already have it in Form 3X so maybe we should extract it to it's own definition and import it so we don't have it defined in multiple places?

toddlees: moved it to the report serializer

toddlees:
> This looks good. I'm a little concerned with how widespread our changes are whenever we need to add a new form type and how we have to scrounge through the whole api looking for places where we need to update it to include the new form. I almost wish we had a centralized file that handled form utils but I feel like that would be cumbersome and not very effective. No real proposed solutions but looking at this PR I would be hard pressed to say whether or not everything that needs to be updated with the new form type has been updated.

@sasha-dresden It concerned me too.  That's partially why we're doing this "tracer bullet" to identify pain points like that without having to be down to the wire implementing the feature for a deadline.  One thing that seems like it would help is to abstract the parts in the external modules into the reports module.  like the transaction module could reference "reports_that_need_to_be_carried_forward" and let the reports module do the business of defining those reports.

If you have any epiphanies please share.  Luckily since we're doing this early, we can brainstorm

lbeaufort (approved):
Exciting work! 🎉

PR #1418, opened 2025-04-03
feature/2121 - Remove text_code and rename text_code_2 to text_code

sasha-dresden:
Issue [FECFILE-2121](https://fecgov.atlassian.net/browse/FECFILE-2121)


django-backend/fecfiler/reports/migrations/0014_form99_swap_text_code.py
danguyf, line 1: Is there a way to do this across two migrations that would be reversible and therefore not need the IgnroeMigration()?  (Answering this question is one of my Innovation Sprint ideas.)

danguyf (approved):
This follows our current process.

lbeaufort (approved):
Thanks @sasha-dresden!

PR #1438, opened 2025-04-14
Patch - Prevent reference before assignment error in dotFEC creation

Elaine-Krauss-TCG:
The create dotfec task raises a value error if there's neither a webprint submission nor an upload submission

Ticket link: https://fecgov.atlassian.net/browse/FECFILE-2153
  
Related PRs:

### DIFF
--- a/django-backend/fecfiler/web_services/tasks.py
+++ b/django-backend/fecfiler/web_services/tasks.py
@@ -61,12 +61,17 @@ def create_dot_fec(
     force_write_to_disk=False,
     file_name=None,
 ):
+    submission = None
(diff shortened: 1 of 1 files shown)


lbeaufort (changes requested):
@Elaine-Krauss-TCG can you link to the ticket?

danguyf (approved):
Tested locally and seeing the new exception in the logs rather than the old error, which is helpful.

lbeaufort (approved):
We tried to collect more data locally, but moving to better exception logging in dev so we can get more info

PR #1451, opened 2025-04-22
efficient aggregate trigger

toddlees:
Ticket link: https://fecgov.atlassian.net/browse/FECFILE-2152


lbeaufort (approved):
Thanks @toddlees! Confirmed with Todd that this was able to finish his local with a copy of the dev db in under a minute.

PR #1469, opened 2025-04-28
feature/2192 - Add request_id to PL/SQL logging statements

dheitzer:
Ticket link: https://fecgov.atlassian.net/browse/FECFILE-2192
  
Related PRs:

### DIFF
--- a/django-backend/fecfiler/settings/base.py
+++ b/django-backend/fecfiler/settings/base.py
@@ -28,7 +28,9 @@ KEY_VALUE = "KEY_VALUE"
 APPEND_SLASH = False
 
 LOG_FORMAT = env.get_credential("LOG_FORMAT", LINE)
-ENABLE_PL_SQL_LOGGING = env.get_credential("ENABLE_PL_SQL_LOGGING", False)
+ENABLE_PL_SQL_LOGGING = get_boolean_from_string(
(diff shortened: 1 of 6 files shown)


django-backend/fecfiler/settings/base.py
```diff
@@ -28,6 +28,9 @@
 APPEND_SLASH = False
 
 LOG_FORMAT = env.get_credential("LOG_FORMAT", LINE)
+ENABLE_PL_SQL_LOGGING = get_boolean_from_string(
```
lbeaufort, line 31: Is it overkill to add a comment here saying something like "Never set this to true for dev/stage/test/prod"?

lbeaufort (approved):
This looks great, @dheitzer! I tested setting `ENABLE_PL_SQL_LOGGING: True` which worked great, and confirmed that `ENABLE_PL_SQL_LOGGING: ` (blank) correctly is interpreted as `False`

One small comment: https://github.com/fecgov/fecfile-web-api/pull/1469#pullrequestreview-2804342660 (resolved)

django-backend/fecfiler/settings/base.py
lbeaufort: Perfect, thank you!

danguyf (approved):

toddlees (changes requested):

PR #1480, opened 2025-05-07
[Sprint 64] Feature/2105 - Schedule F Aggregation

Elaine-Krauss-TCG:
Ticket link: [here](https://fecgov.atlassian.net/browse/FECFILE-2105)
  
Related PRs: [API](https://github.com/fecgov/fecfile-web-api/pull/1480), [App](https://github.com/fecgov/fecfile-web-app/pull/2719)


django-backend/fecfiler/transactions/views.py
```diff
@@ -396,6 +438,26 @@ def save_transaction(self, transaction_data, request):
         delete_carried_forward_loans_if_needed(transaction_instance, committee_id)
         delete_carried_forward_debts_if_needed(transaction_instance, committee_id)
 
+        # Handle changing general_election_year for Schedule F transactions
+        if (
```
toddlees, line 442: save_transaction is getting pretty big.  I know it would mean passing in more info to ```recalculate_schedule_f_aggregates```, but if we did, could we not figure out ALL of the transactions that need to be recalculated in one place rather than spread across 3 spots in save_transaction?

django-backend/fecfiler/transactions/views.py
```diff
@@ -508,6 +582,50 @@ def remove_transaction_from_report(self, request):
...
+            transaction.schedule_f.general_election_year
+        )
+
+        to_update = shared_entity_transactions.filter(
+            Q(id=transaction.id)
+            | Q(date__gt=transaction.date)
+            | Q(Q(date=transaction.date) & Q(created__gt=transaction.created))
+        ).order_by("date", "created")
+
+        previous_transaction = previous_transactions.first()
```
toddlees, line 613: i believe this snippet is duplicate work

toddlees (changes requested):

django-backend/fecfiler/transactions/utils.py
```diff
@@ -37,3 +38,48 @@ def add_location(data, contact: Contact, identifier: str):
...
+            Q(contact_2__candidate_office=office),
+            Q(contact_2__candidate_state=state),
+            Q(contact_2__candidate_district=district),
+        )
+
+    original_transaction = None
+    if transaction_id:
+        original_transaction = queryset.get(id=transaction_id)
+        query = query.filter(
+            Q(created__lt=original_transaction.created) | ~Q(date=date)
```
dheitzer, line 81: should this be ~Q(date >= date) to make sure we're always getting previous transactions? Also, couldn't transactions with older created dates still have newer dates?

django-backend/fecfiler/transactions/views.py
```diff
@@ -509,6 +672,59 @@ def remove_transaction_from_report(self, request):
         transaction.reports.remove(report)
         return Response("Transaction removed from report")
 
+    def process_general_election_year_aggregates(self, transaction_instance):
```
dheitzer, line 675: I'm wondering if we can name this function more descriptive - maybe update_previous_transactions_aggregates()?

django-backend/fecfiler/transactions/views.py
```diff
@@ -509,6 +672,59 @@ def remove_transaction_from_report(self, request):
...
+                    previous_transaction.schedule_f.aggregate_general_elec_expended
+                )
+
+            trans.schedule_f.aggregate_general_elec_expended = (
+                trans.schedule_f.expenditure_amount + previous_aggregate
+            )
+            updated_schedule_fs.append(trans.schedule_f)
+            previous_transaction = trans
+
+        ScheduleF.objects.bulk_update(
```
dheitzer, line 723: I'm wondering if we might want to set the batch_size parameter in case we end up with an excessive number of previous transactions.  Thoughts?
https://docs.djangoproject.com/en/5.2/ref/models/querysets/#bulk-update

django-backend/fecfiler/transactions/views.py
```diff
@@ -509,6 +672,59 @@ def remove_transaction_from_report(self, request):
...
+            transaction.id,
+            transaction.date,
+            transaction.aggregation_group,
+            None,
+            transaction.contact_2.id,
+            None,
+            transaction.schedule_f.general_election_year
+        )
+
+        to_update = shared_entity_transactions.filter(
```
dheitzer, line 702: Should this be including submitted transactions and such?

django-backend/fecfiler/transactions/views.py
Elaine-Krauss-TCG: It might be unnecessary because we're just altering one column, but I'm open to the idea.  Any thoughts on what a good size for the batch size would be?
image

django-backend/fecfiler/transactions/utils.py
```diff
@@ -37,3 +38,48 @@ def add_location(data, contact: Contact, identifier: str):
     data[f"{identifier}_city"] = contact.city
     data[f"{identifier}_state"] = contact.state
     data[f"{identifier}_zip"] = contact.zip
+
+
+def filter_for_previous_transactions(
```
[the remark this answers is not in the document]
Elaine-Krauss-TCG: Sounds good; I'm working on a commit for that

django-backend/fecfiler/transactions/utils.py
Elaine-Krauss-TCG: I'm pretty convinced that I've got all the logic working correctly, and I think I have a pretty comprehensive set of unit and E2E tests, but if you see even a chance that there's room for an edge-case, then we should definitely take that seriously.  I'll take another look at my unit tests and make sure that there are tests for the aggregation logic around creation order

toddlees: I think it's correct @Elaine-Krauss-TCG maybe we can just add a comment explaining the purpose of this one filter.  I think the confusion is in thinking it's meant to be the whole filter, when really it's just addressing a subset of the transactions that are filtered in the whole function (ones that are the same date but created before the original)

toddlees (approved):
Anything you can address in the comments is great. Overall, this is in a good place

django-backend/fecfiler/transactions/utils.py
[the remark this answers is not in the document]
Elaine-Krauss-TCG: I broke it down a lot, and hopefully between comments and the names of helper functions, it should be a lot easier to read through.  Additionally, by having helper functions return Q objects, I was able to get the newly renamed `filter_queryset_for_previous_transactions_in_aggregation()` to only make one `.filter()` call

danguyf (approved):
(chef's kiss emoji)

dheitzer (approved):

django-backend/fecfiler/transactions/utils.py
dheitzer: yes @toddlees I think I misunderstood this filter - thanks for all the checking and clarification @toddlees @Elaine-Krauss-TCG

PR #1489, opened 2025-05-13
Update python to version 3.11

lbeaufort:
Ticket link: https://fecgov.atlassian.net/browse/FECFILE-2037

- Update to python 3.11
- Remove old readme
- Update flake8 to latest
- Fix/noqa some linting errors
- Update Python orb version

Related PRs: https://github.com/fecgov/fecfile-web-app/pull/2731


dheitzer (approved):
looks nice! great to upgrade flake8 finally too.

django-backend/README.md
```diff
@@ -1,23 +0,0 @@
-# NXG-FEC-Backend
```
toddlees, line 1: nice!

django-backend/fecfiler/transactions/migrations/0020_trigger_save_on_transactions.py
```diff
@@ -4,71 +4,72 @@
...
-
-    dependencies = [
-        ("transactions", "0019_aggregate_committee_controls"),
-    ]
+def trigger_save_on_transactions(apps, schema_editor):
+    transactions = apps.get_model("transactions", "transaction")
+    committees = apps.get_model("committee_accounts", "committeeaccount")
+    contacts = apps.get_model("contacts", "contact")
 
-    def trigger_save_on_transactions(apps, schema_editor):
```
toddlees, line 13: I was wondering how this worked but now it makes sense. We're never trying to call it from an instance of Migration like `migration.trigger_save_on_transaction()` or `self.trigger_save_on_transaction()`.  So it's being defined in a class where it SHOULD be a class method, but it's never being used as one, so the missing `self` is not an issue

toddlees (approved):

PR #1515, opened 2025-06-09
feature/2133 - Update to use 8.4 spec only where needed

sasha-dresden:
Issue [FECFILE-2133](https://fecgov.atlassian.net/browse/FECFILE-2133)

Validate: [PR356](https://github.com/fecgov/fecfile-validate/pull/356)
APP: [PR2818](https://github.com/fecgov/fecfile-web-app/pull/2818)


toddlees (approved):
Addressed some issues.  When I added the unit tests it showed that we needed to change the override values.  Plus the schema used for col sequence in the .fec serializer is actually a different one from the specific transaction type (I had forgotten this detail).  I needed to update the c2 override schema on the validator (filename, and removing extra fields).  moved override into a shared utility but removed the instance from the view serializer because it didn't apply there.

PR #1526, opened 2025-06-18
[Sprint 65] feature/1670 - Add name field to Form 24

sasha-dresden:
Issue [FECFILE-1670](https://fecgov.atlassian.net/browse/FECFILE-1670)

APP PR: [PR2845](https://github.com/fecgov/fecfile-web-app/pull/2845)

### DIFF
--- a/django-backend/fecfiler/reports/form_24/models.py
+++ b/django-backend/fecfiler/reports/form_24/models.py
@@ -18,7 +18,7 @@ class Form24(Model):
 
     report_type_24_48 = TextField(null=True, blank=True)
     original_amendment_date = DateField(null=True, blank=True)
-    name = TextField(null=False, blank=False, unique=True)
+    name = TextField(null=False, blank=False)
 
     class Meta:
         app_label = "reports"
--- a/django-backend/fecfiler/reports/form_24/models.py
+++ b/django-backend/fecfiler/reports/form_24/models.py
@@ -1,14 +1,14 @@
 import uuid
-from django.db import models
+from django.db.models import Model, UUIDField, TextField, DateField
 import structlog
 
 logger = structlog.get_logger(__name__)
 
 
-class Form24(models.Model):
+class Form24(Model):
     """Generated model from json schema"""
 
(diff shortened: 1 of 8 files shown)


Elaine-Krauss-TCG (dismissed):

django-backend/fecfiler/reports/migrations/00018_form24_name.py
```diff
@@ -0,0 +1,36 @@
...
+        ("reports", "0017_form99_filing_frequency_form99_pdf_attachment"),
+    ]
+
+    operations = [
+        IgnoreMigration(),
+        migrations.AddField(
+            model_name="form24",
+            name="name",
+            field=models.TextField(null=False, blank=False, unique=True),
+        ),
```
Elaine-Krauss-TCG, line 34: This part surprisingly fails when you run it while already having multiple saved F24 reports.  It doesn't like that the new column is created containing null values.  You'll probably want to make the field as null=True, set the name values with update_form24_names, and then modify the field to null=False

django-backend/fecfiler/reports/form_24/models.py
```diff
@@ -1,23 +1,24 @@
...
         primary_key=True,
         serialize=False,
         unique=True,
     )
 
-    report_type_24_48 = models.TextField(null=True, blank=True)
-    original_amendment_date = models.DateField(null=True, blank=True)
+    report_type_24_48 = TextField(null=True, blank=True)
+    original_amendment_date = DateField(null=True, blank=True)
+    name = TextField(null=False, blank=False, unique=True)
```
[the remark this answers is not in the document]
sasha-dresden: Ah good point. It needs to be unique to a committee. I updated the code so it's no longer unique at the sql level but instead performs a check at the form24 serializer

django-backend/fecfiler/reports/migrations/00018_form24_name.py
sasha-dresden: Ah good point. Got that fixed.

PR #1533, opened 2025-06-26
Feature/2105c - Schedule F Aggregation

Elaine-Krauss-TCG:
Ticket link: [here](https://fecgov.atlassian.net/browse/FECFILE-2105), also fixes [FECFILE-2323](https://fecgov.atlassian.net/browse/FECFILE-2323)

Related PRs: [App](https://github.com/fecgov/fecfile-web-app/pull/2852), [API](https://github.com/fecgov/fecfile-web-api/pull/1533)

### DIFF
--- a/django-backend/fecfiler/transactions/views.py
+++ b/django-backend/fecfiler/transactions/views.py
@@ -182,15 +182,12 @@ class TransactionViewSet(CommitteeOwnedViewMixin, ModelViewSet):
             aggregation_group = request.query_params["aggregation_group"]
             assert (
                 date
+                and contact_1_id
                 and aggregation_group
             )  # Raises error if any value is not truthy (i.e, "" or None)
         except Exception:
-            message = "date and aggregate_group are required params"
+            message = "contact_1_id, date, and aggregate_group are required params"
             return Response(message, status=status.HTTP_400_BAD_REQUEST)
-
-        if not contact_1_id:
-            message = "no contact id provided"
-            return Response(message, status=status.HTTP_404_NOT_FOUND)
 
         return self.get_previous(
             self.get_queryset(),
@@ -219,7 +216,6 @@ class TransactionViewSet(CommitteeOwnedViewMixin, ModelViewSet):
 
             assert (
                 date
-                and office
                 and aggregation_group
                 and election_code
             )  # Raises error if any value is not truthy (i.e, "" or None)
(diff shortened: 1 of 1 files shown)


django-backend/fecfiler/contacts/views.py
```diff
@@ -162,13 +162,16 @@ def candidate_lookup(self, request):
...
+            response = requests.get(
+                FEC_API_CANDIDATE_LOOKUP_ENDPOINT, params=params
+            )
+            if response.status_code != status.HTTP_404_NOT_FOUND:
+                response.raise_for_status()
 
-        json_results = response.json()
+            json_results = response.json()
+        else:
+            json_results = {}
```
Elaine-Krauss-TCG, line 174: These were the source of the Contact Lookup bug from FECFILE-2323.  Essentially, if the query was less than 3 characters long, then the FEC API would respond with a 422 error.  `raise_for_status()` would then throw an error, which would interrupt the rest of the function.  This prevented our API from returning _any_ contacts on Committee and Candidate lookup if the search query was not at least 3 characters long.

Elaine-Krauss-TCG: (Another valid approach here would have been to add the 422 error to the status codes ignored, but I figured it would be a good idea to avoid throwing requests at the FEC API that we already know will fail)

django-backend/fecfiler/transactions/views.py
```diff
@@ -180,6 +180,11 @@ def previous_transaction_by_entity(self, request):
             contact_1_id = request.query_params["contact_1_id"]
             date = request.query_params["date"]
             aggregation_group = request.query_params["aggregation_group"]
+            assert (
+                date
+                and contact_1_id
+                and aggregation_group
+            )  # Raises error if any value is not truthy (i.e, "" or None)
```
Elaine-Krauss-TCG, line 187: The source of the bug, as I found it, was that when you make a request to any of our previous transaction endpoints, if you included a required parameter's name in the url but did not include a _value_, then it wasn't being caught here.  This would lead to weird and inconsistent behavior, including letting `previous_transaction_by_payee_candidate` match against any schedule F transaction if no value was provided for contact_2_id.  I'll provide an example below:

```
https://dev-api.fecfile.fec.gov/api/v1/transactions/previous/payee-candidate/?transaction_id=&contact_2_id=&date=2025-06-30&aggregation_group=COORDINATED_PARTY_EXPENDITURES&general_election_year=2024
```

and we can see in this screenshot that it was resulting in a 200 response when it obviously should have gotten a 400 response
image

sasha-dresden (approved):

django-backend/fecfiler/transactions/views.py
toddlees: the symptom of this is if you make a schedule F with a new candidate contact, your aggregate doesn't get calculated
image

toddlees:
Reverted last 404 change.  We don't need contact to act differently on the api to solve the bug noted above.  we needed to update the call from the app.

PR #1538, opened 2025-06-30
Patch: Set Celery storage to disk in Dockerfile

lbeaufort:
Ticket link: N/A

Set Celery storage to disk in Dockerfile. Before this change, local Celery storage was defaulting to S3
  
Test by generating a .fec file locally


toddlees (approved):
Exactly what I wanted to do!

PR #1544, opened 2025-07-01
feature/2287 - Create submit script for locust

dheitzer:
Ticket link: https://fecgov.atlassian.net/browse/FECFILE-2287
  
Related PRs:

Currently, the new submit workflow load testing works as follows:

1. during `on_start` (before load tests are run), the following api interactions occur to prepare the reports created for the load test run for the submission load test as would be initiated by the web client during this flow.  For each report created for the load tests:
    - the `calculate-summary` task is posted
    - the report is retrieved from the API up to 3 times looking for calculated summary values
    - the report is updated in the API with confirmation email
    - the updated report info is retrieved from the API
2. the `submit_reports` task is executed by whatever number of users (threads) is configured for the test run.  The following api interactions occur for each prepared report just once* as would be initiated by the web client during this flow:
    - the report is updated in the API with the `treasurer_last_name` and `treasurer_first_name`
    - the report_id is posted to the submit endpoint with
    - the report is then retrieved in a polling fashion up to 3 times (waiting in between the poll length used by the API to check FEC) checking for an fec_status of "ACCEPTED".  If this is not received, the test is failed.  If it is received, it is success.

*Note: The `submit_reports` locust test uses a subset of report ids to ensure that each report_id prepared for locust is only submitted once by a single thread.  This test task stops running when all report_ids have been submitted once.


sasha-dresden (approved):

lbeaufort (approved):
This looks great, thanks @dheitzer!

PR #1564, opened 2025-07-16
[WIP] FECFILE-2061: logging update

danguyf:
Ticket link:
  
Related PRs:
None (as yet)

There are a few things going on here:
2.) The logging was updated as requested in the ticket.
3.) Removed the DotFECSubmitter and WebPrintSubitter ABCs.

That last one may get me in trouble and I'm completely open to the possibility that I've taken a wrong turn.  My thinking is that the mock versions did nothing but return some JSON -- and often the same JSON regardless of the method, so why not instead have a class that returns different JSON payloads on one hand and on the other hand always use the "real" submitter but have it differentiate between whether the response object should get a mock response or a real one?

There are a few bits that I would like to refactor further but that could be a follow-up ticket.


danguyf:
Closing being I refactored too close to the sun.

PR #1566, opened 2025-07-16
FECFILE-2061: logging update

danguyf:
Ticket link:

Related PRs:
None (as yet)

There are a few things going on here:
2.) The logging was updated as requested in the ticket.
3.) Removed the DotFECSubmitter ABC.

That last one may get me in trouble and I'm completely open to the possibility that I've taken a wrong turn. My thinking is that many mock versions simply return some JSON -- and often the same JSON regardless of the method, so why not instead have a class that returns different JSON payloads on one hand and on the other hand always use the "real" submitter but have it differentiate between whether the response object should get a mock response or a real one?

There are a few bits that I would like to refactor further but that could be a follow-up ticket.

### DIFF
--- a/django-backend/fecfiler/web_services/tasks.py
+++ b/django-backend/fecfiler/web_services/tasks.py
@@ -12,7 +12,6 @@ from fecfiler.web_services.models import (
 from fecfiler.web_services.dot_fec.dot_fec_composer import compose_dot_fec
 from fecfiler.web_services.dot_fec.dot_fec_submitter import (
     EFODotFECSubmitter,
-    MockDotFECSubmitter,
 )
 from fecfiler.web_services.dot_fec.web_print_submitter import (
     EFOWebPrintSubmitter,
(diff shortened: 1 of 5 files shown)


django-backend/fecfiler/web_services/tasks.py
```diff
@@ -154,7 +150,7 @@ def submit_to_fec(
 
     """Submit to FEC"""
     try:
-        submission_type_key = DOT_FEC_KEY if not mock else MOCK_DOT_FEC_KEY
+        submission_type_key = EFO_SUBMITTER_KEY
```
toddlees, line 153: This breaks the functionality of being able to call the task and decide whether it will be mocked or not.  I get the goal of reducing how much code is different between the mock and the real submitter.  Maybe a way to do that without making the the submitter and print submitter tied to the environment variable is to just pull the logging out of the class so all that's left is the different part

PR #1571, opened 2025-07-17
FECFILE-2061: logging update

danguyf:
Ticket link:

Related PRs:
None (as yet)

Note:
2.) Additional logging, aligned to the request in the ticket, was added to the `save_fec_response` methods of dot_fec_submitter and web_print_submitter; the existing debug logging in those classes was left in place for debugging.

### DIFF
--- a/django-backend/fecfiler/web_services/models.py
+++ b/django-backend/fecfiler/web_services/models.py
@@ -234,7 +234,8 @@ class WebPrintSubmission(BaseSubmission):
             fec_response_json = json.loads(response_string)
         except Exception as error:
             logger.error(
-                "Failed to parse JSON response from web print submission: {response_string}"
+                "Failed to parse JSON response from web print submission: "
+                f"{response_string}"
             )
             raise error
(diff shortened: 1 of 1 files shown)


django-backend/fecfiler/web_services/models.py
```diff
@@ -197,6 +197,11 @@ def save_fec_response(self, response_string):
             logger.error("Failed to parse JSON response from upload submission")
             raise error
 
+        if fec_response_json.get("status") != FECStatus.ACCEPTED.value:
```
toddlees, line 200: we don't want to throw an error when the status is FECStatus.PENDING.value

toddlees (changes requested):

django-backend/fecfiler/web_services/models.py
danguyf: Addressed FECStatus.PROCESSING and that FECStatus.COMPLETED should also not be sent to error log.

PR #1609, opened 2025-08-08
feature/2396 - [Bug] Back to FECfile+ button does not work in login.gov

dheitzer:
Ticket link: https://fecgov.atlassian.net/browse/FECFILE-2396
  
Related PRs:

Since the front-end uses the presence of the `ffapi_timeout` cookie to determine whether the user is authenticated, we should only set it on the backend if the user is authenticated.


danguyf:
Working for me in DEV post-deploy!

PR #1640, opened 2025-08-20
feature/2377 -  Create management command for granting access in case of account lockout

dheitzer:
Ticket link: https://fecgov.atlassian.net/browse/FECFILE-2377
  
Related PRs:

New command plus refactored out account creation logic and updated view accordingly so it's shared with both.


toddlees (approved):
Works like a charm.  updated tag for opensearch to Account Override

PR #1716, opened 2025-09-25
feature/2012 - Update query_fec_api to include user agent and update all uses of fec_api to go through query_fec_api function

sasha-dresden:
Issue [FECFILE-2012](https://fecgov.atlassian.net/browse/FECFILE-2012)

### DIFF
--- a/django-backend/fecfiler/committee_accounts/utils.py
+++ b/django-backend/fecfiler/committee_accounts/utils.py
@@ -2,7 +2,7 @@ import requests
 import re
 from rest_framework.exceptions import ValidationError
 from .models import CommitteeAccount, Membership
-
+from rest_framework.status import HTTP_404_NOT_FOUND
 from fecfiler import settings
 import redis
 import json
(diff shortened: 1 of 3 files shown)


danguyf (approved):
Effective and good refactor

django-backend/fecfiler/committee_accounts/utils.py
```diff
@@ -105,14 +105,29 @@ def get_committee_account_data(committee_id):
 """
 
 
-def query_fec_api(endpoint, params):
+def get_environment():
```
dheitzer, line 108: I'm wondering if we can add a dedicated env var for this to have better control.  I feel like it could be easy to miss updating this if we changed/got rid of the timeout cookie.

dheitzer (changes requested):
just a few thoughts

django-backend/fecfiler/committee_accounts/utils.py
```diff
@@ -105,14 +105,29 @@ def get_committee_account_data(committee_id):
...
+    ffapi = settings.FFAPI_TIMEOUT_COOKIE_NAME
+    prefix = "ffapi_timeout_"
+    return ffapi.replace(prefix, "")
```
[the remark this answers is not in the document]
sasha-dresden: Oh I just realized this is a different concern. My previous comment was for the one above. For this one, I don't think we could have a NONE value here since settings.FFAPI_TIMEOUT_COOKIE_NAME raises an exception if it isn't defined due to this line:

if not FFAPI_TIMEOUT_COOKIE_NAME:
    raise Exception("FFAPI_TIMEOUT_COOKIE_NAME is not set!")

django-backend/fecfiler/committee_accounts/utils.py
```diff
@@ -105,14 +105,23 @@ def get_committee_account_data(committee_id):
 """
 
 
-def query_fec_api(endpoint, params):
+def query_fec_api_single(endpoint, params):
```
toddlees, line 108: EXCELLENT job refactoring this!  You improved the code using it and the readability.  I especially like renaming `json_results` to `fec_api_candidates`.  It captures what the variable is.

Since this is now used in both the committee account module and the contact module, let's move it to the shared module.

toddlees (changes requested):
Great refactor.  I see how parsing the timeout cookie doesn't require changing environment variables.  For this case let's make the improvement now and replace the timeout cookie name variable with the environment (maybe let's call it `space` to match the cloud.gov language) variable

dheitzer (approved):

toddlees (approved):

PR #1750, opened 2025-10-17
Custom list logic for members to always place current user first for 1st page

sasha-dresden:
Issue [FECFILE-2402](https://fecgov.atlassian.net/browse/FECFILE-2402)

APP [PR3140](https://github.com/fecgov/fecfile-web-app/pull/3140)

### DIFF
--- a/django-backend/fecfiler/committee_accounts/views.py
+++ b/django-backend/fecfiler/committee_accounts/views.py
@@ -42,7 +42,7 @@ class CommitteeViewSet(viewsets.GenericViewSet, mixins.ListModelMixin):
 
     def get_queryset(self):
         user = self.request.user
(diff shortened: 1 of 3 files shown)


django-backend/fecfiler/committee_accounts/views.py
```diff
@@ -301,3 +301,33 @@ def update(self, request, *args, **kwargs):
         )
 
         return super().update(request, *args, **kwargs)
+
+    def list(self, request, *args, **kwargs):
+        you_first = request.query_params.get("you_first")
+        logger.info(you_first)
```
Elaine-Krauss-TCG: Was this a debugging aid, or do we want to log it long-term?  I'm fine either way, though.

PR #1770, opened 2025-11-18
Update dev wiki

max-zaremba-tcg:
Removed detailed setup instructions for Docker and environment variables, and added links to the wiki for setup and testing instructions.

Ticket link: [FECFILE-2642](https://fecgov.atlassian.net/browse/FECFILE-2641)


README.md
danguyf: With your above addition, all of this can be removed.

PR #1823, opened 2025-12-17
feature/2683 - Update C2 update_in_future_reports logic to prevent parent_transaction overwrite

sasha-dresden:
Issue [FECFILE-2683](https://fecgov.atlassian.net/browse/FECFILE-2683)


toddlees (dismissed):
We DEFINITELY want unit tests and or e2e for this scenario.  If i don't get back to this this week, we can just create a follow up ticket for tests and unblock this

PR #1837, opened 2026-01-06
Feature/2704 - Create streamlined transaction serializer for list endpoint

sasha-dresden:
Issue [FECFILE-2704](https://fecgov.atlassian.net/browse/FECFILE-2704)
APP PR [PR3397](https://github.com/fecgov/fecfile-web-app/pull/3397)

These changes are to create and use a minimal transaction object to be used when querying for a list of transactions (specifically in transaction list page and the contact page) so that we do not have to query/retrieve an entire transaction object for these scenarios.


danguyf (approved):
Of a piece with the related web-app PR.

PR #1874, opened 2026-01-29
feature/2547 - Optimize /previous/ endpoints database usage

sasha-dresden:
Issue [FECFILE-2547](https://fecgov.atlassian.net/browse/FECFILE-2547)

APP [PR3496](https://github.com/fecgov/fecfile-web-app/pull/3496)

List of changes

- Add secondary queryset for previous transactions which only grabs info needed (don't need Reports for example)
- Cleanup ordering of TransactionManager so all clauses come after the querysets instead of some before and some after
- Cleanup ordering of TransactionViewSet so all utility functions come after the actions
- Update get_previous logic to use new queryset and tuple comparison for the date check.
- Update returned value from get_previous to simply return the three aggregate value numbers instead of the entire serialized Transaction
- Updated tests to test for correct values returned rather than checking the id of the transaction returned.

### DIFF
--- a/django-backend/fecfiler/transactions/views.py
+++ b/django-backend/fecfiler/transactions/views.py
@@ -251,7 +251,7 @@ class TransactionViewSet(CommitteeOwnedViewMixin, ModelViewSet):
         return Response("Transaction removed from report")
 
     @action(detail=False, methods=["get"], url_path=r"previous/entity")
-    def previous_transaction_by_entity(self, request):
+    def previous_aggregate_by_entity(self, request):
         try:
             contact_1_id = request.query_params["contact_1_id"]
             assert (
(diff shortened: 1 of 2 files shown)


django-backend/fecfiler/transactions/views.py
```diff
@@ -198,184 +218,217 @@ def update_itemization_aggregation(self, request, pk=None):
...
         )
+        if state:
+            filters &= Q(contact_2__candidate_state=state)
+        if district:
+            filters &= Q(contact_2__candidate_district=district)
+
+        return self.find_and_process_previous(request, filters)
 
     @action(detail=False, methods=["get"], url_path=r"previous/payee-candidate")
     def previous_transaction_by_payee_candidate(self, request):
```
toddlees: I like that you're making the return data specific.  We're not really returning a transaction anymore.  Maybe it's worth calling these endpoints `previous-aggregate/*`

sasha-dresden: Oh yea, I suppose that is true haha. Good call. Renamed them

PR #1974, opened 2026-04-06
FECFILE-3016: remove sphinx/gh-pages, docs

danguyf:
Ticket link:
https://fecgov.atlassian.net/browse/FECFILE-3016

Related PRs:
https://github.com/fecgov/fecfile-validate/pull/430

Testing/validation thoughts [in the ticket](https://fecgov.atlassian.net/browse/FECFILE-3016?focusedCommentId=56405).


dheitzer:
@danguyf looks good.  One thing I'm not clear about though even after digging - are we using Sphinx for fecfile-validate? We still  [have it in the requirements.txt](https://github.com/fecgov/fecfile-validate/blob/65d958383dbfe7eb965384eabf582a172c56712c/requirements.txt#L1) and I'm wondering if it can be removed?

danguyf:
> @danguyf looks good. One thing I'm not clear about though even after digging - are we using Sphinx for fecfile-validate? We still [have it in the requirements.txt](https://github.com/fecgov/fecfile-validate/blob/65d958383dbfe7eb965384eabf582a172c56712c/requirements.txt#L1) and I'm wondering if it can be removed?

Good catch.  I don't see where we're using `sphinx` or `rfc3339-validator` in `fecfile-validate` so maybe we need a ticket to clean up its `requirements.txt`? (cc: @toddlees)

danguyf:
> Good catch. I don't see where we're using `sphinx` or `rfc3339-validator` in `fecfile-validate` so maybe we need a ticket to clean up its `requirements.txt`? (cc: @toddlees)

 `rfc3339-validator`  is used by `jsonschema`.

toddlees (approved):

PR #2015, opened 2026-04-28
[WIP - do not merge] Feature/338 Serve API doc static assets 

dheitzer:
Ticket link: https://fecgov.atlassian.net/browse/FECFILE-338
  
Related PRs:

PR #2017, opened 2026-04-30
feature/338 - use drf spectacular sidecar for swagger

dheitzer:
Ticket link: https://fecgov.atlassian.net/browse/FECFILE-338
  
Related PRs: We implemented [an approach](https://github.com/fecgov/fecfile-web-api/pull/2015) where we host the .js and .css static assets in our codebase.  This is not ideal however since we want do not want to take on management of the asset for upgrade/security purposes.

Confirmed that static swagger assets are being cached by Cloudfront:
 
image


lbeaufort (commented):
@dheitzer can you clean up the commit history here to remove the dev testing commits?

PR #2043, opened 2026-05-18
feature/2996 - Update aggregate so only runs if aggregate_group

sasha-dresden:
Issue [FECFILE-2996](https://fecgov.atlassian.net/browse/FECFILE-2996)


danguyf (approved):
This makes sense.

toddlees (approved):
looking good

PR #2044, opened 2026-05-19
Feature/3102 dotfec divide exact percentage by 100

Elaine-Krauss-TCG:
Ticket link: https://fecgov.atlassian.net/browse/FECFILE-3102
  
Related PRs:


django-backend/fecfiler/web_services/dot_fec/dot_fec_serializer.py
```diff
@@ -71,6 +71,22 @@ def text_to_date_serializer(model_instance, field_name, mapping):
...
+    if schedule:
+        interest_rate = schedule.loan_interest_rate
+        is_percent = schedule.loan_interest_rate_is_percent
+
+        if not is_percent:
+            return interest_rate
+        else:
+            try:
+                no_percent_symbol = str(interest_rate).strip('%')
+                return str(float(no_percent_symbol) / 100)
```
danguyf: I think this might be better as `str(Decimal(no_percent_symbol) / 100)` to avoid potential floating-point precision issues.

django-backend/fecfiler/web_services/dot_fec/tests/test_dot_fec_serializer.py
```diff
@@ -198,3 +206,138 @@ def test_get_value_from_path(self):
...
+
+    def test_serialize_loan_interest_rate_user_defined(self):
+        contact_org = create_test_organization_contact(
+            "Organization",
+            self.committee.id
+        )
+        loan = create_loan(
+            self.committee,
+            contact_org,
+            24000,
```
danguyf: SonarCloud would stop complaining if this was a decimal but I don't think that's a big deal.

danguyf (approved):
This works!  I have two suggestions.

django-backend/fecfiler/transactions/migrations/0004_schedulec_loan_due_date_is_date_and_more.py
```diff
@@ -0,0 +1,59 @@
...
+    ScheduleC1 = apps.get_model(  # noqa: N806
+        "transactions.ScheduleC1"
+    )
+
+    for Schedule in [ScheduleC, ScheduleC1]:
+        instances = list(Schedule.objects.all())
+        for instance in instances:
+            interest_rate = instance.loan_interest_rate
+            try:
+                float(str(interest_rate).strip('%'))
```
toddlees: I don't think we confidently make this conclusion.  a non-exact-percent entry can still be "1%" just entered in the other mode.  In that case we would not want to transform it in the .fec generation.  I think the safest thing for us to do is to set all the ones where we don't know which option they picked to be False.

Happy to discuss though

django-backend/fecfiler/web_services/dot_fec/dot_fec_serializer.py
```diff
@@ -71,6 +71,22 @@ def text_to_date_serializer(model_instance, field_name, mapping):
...
+    schedule = model_instance.schedule_c or model_instance.schedule_c1
+    if schedule:
+        interest_rate = schedule.loan_interest_rate
+        is_percent = schedule.loan_interest_rate_is_percent
+
+        if not is_percent:
+            return interest_rate
+        else:
+            try:
+                no_percent_symbol = str(interest_rate).strip('%')
```
toddlees: This is looking great!  Is the front end not going to be sending `25` after the new ui moves the % out of the input field like:`'25'%`?  we might not need to do this strip

django-backend/fecfiler/web_services/dot_fec/dot_fec_serializer.py
toddlees: agreed.  precision is important here

django-backend/fecfiler/web_services/dot_fec/dot_fec_serializer.py
```diff
@@ -71,6 +71,22 @@ def text_to_date_serializer(model_instance, field_name, mapping):
...
+                return str(float(no_percent_symbol) / 100)
+            except ValueError:
+                raise ValueError("Interest rate is not a valid number")
```
toddlees: I like this clarity.  Maybe we can even catch when `schedule` doesn't exist or the fields don't exist

toddlees (changes requested):

django-backend/fecfiler/web_services/dot_fec/dot_fec_serializer.py
Elaine-Krauss-TCG: Ah, is that so?  Easy change, then

django-backend/fecfiler/web_services/dot_fec/dot_fec_serializer.py
Elaine-Krauss-TCG: An additional validation error was added for the case where it attempts to serialize loan_interest_rate on a transaction without either a schedule c or c1.  Both error messages have been expanded to include the transaction id.

django-backend/fecfiler/web_services/dot_fec/dot_fec_serializer.py
Elaine-Krauss-TCG: The serializer no longer attempts to strip the percentage symbol.

django-backend/fecfiler/transactions/migrations/0004_schedulec_loan_due_date_is_date_and_more.py
Elaine-Krauss-TCG: Removed!  It just sets the field to false for all relevant transactions

toddlees (approved):
This is great!  Really clean work

PR #2045, opened 2026-05-20
Update CommitteeAccount update() to handle name change

sasha-dresden:
Issue [FECFILE-2015](https://fecgov.atlassian.net/browse/FECFILE-2015)

APP [PR3973](https://github.com/fecgov/fecfile-web-app/pull/3973)


toddlees (changes requested):
I'd like for us to not use the committee account view for this.  We have an existing endpoint on users/views.py called update_current that serves the purpose of this ticket.  we can call it from the app the same way we call it on the profile page

sasha-dresden:
Closing. By using the existing path, we don't need to make any modifications to the API

PR #2052, opened 2026-05-26
[Sprint 88] feature/2819 - Setup permission based E2E endpoints

sasha-dresden:
Issue [FECFILE-2819](https://fecgov.atlassian.net/browse/FECFILE-2819)


danguyf (approved):
Slick!

django-backend/fecfiler/settings/permissions.py
```diff
@@ -0,0 +1,10 @@
+from django.conf import settings
+from rest_framework.permissions import BasePermission
+from rest_framework.exceptions import NotFound
+
+
+class IsE2EEnabled(BasePermission):
+    def has_permission(self, request, view):
+        if settings.E2E_TEST:
+            return True
+        raise NotFound()
```
dheitzer: Any reason we're raising an error here instead of just returning `settings.E2E_TEST`? I'm not seeing examples of raised errors in `has_permissions` [here for example](https://www.django-rest-framework.org/api-guide/permissions/#examples).

django-backend/fecfiler/contacts/views.py
```diff
@@ -340,10 +341,13 @@ def get_max_results(self, request):
         )
         return max_fecfile_results, max_fec_results
 
-    @action(detail=False, methods=["post"], url_path="e2e-delete-all-contacts")
+    @action(
+        detail=False,
+        methods=["post"],
+        url_path="e2e-delete-all-contacts",
+        permission_classes=[IsE2EEnabled],
```
dheitzer: Will this return the 404 that was being sent before (as opposed to 403?).  I'm not sure if that's an important distinction or not.

dheitzer (dismissed):
Do we know if we can remove the endpoints if `settings.E2E_TEST` is not set per the ticket comment? 

`It might be more ideal to not include the endpoint in the routing if possible.` 

maybe something in urls.py? For example, we exclude all the mock_oidc_provider on a switch [here](https://github.com/fecgov/fecfile-web-api/blob/f95664df07bbc1224f68389d8423fb6be531569f/django-backend/fecfiler/urls.py#L42)

django-backend/fecfiler/settings/permissions.py
sasha-dresden: Yea, so the reason I used NotFound() is because if you just return False it evaluates to 403 Forbidden, and in this situation we want to hide the existence of the route altogether, so we return NotFound => 404

django-backend/fecfiler/contacts/views.py
sasha-dresden: This goes along with the reason why the permission returns NotFound. By returning NotFound there, we get the same 404 here as before.

sasha-dresden:
> Do we know if we can remove the endpoints if `settings.E2E_TEST` is not set per the ticket comment?
> 
> `It might be more ideal to not include the endpoint in the routing if possible.`
> 
> maybe something in urls.py? For example, we exclude all the mock_oidc_provider on a switch [here](https://github.com/fecgov/fecfile-web-api/blob/f95664df07bbc1224f68389d8423fb6be531569f/django-backend/fecfiler/urls.py#L42)

Yea I had tried to do that originally using the routing urls.py but ran into some issues, but thinking about it more today (and doing a bit more research) I found out you could just wrap the entire action in if block! Apparently django builds it's routes at runtime so if E2E_TEST is False, that action just doesn't exist. However, this does lead to it actually returning a 405. This is because the router says, well, this is a POST request and there is a matching create() method, which we have overridden to always return 405 for reports. (Because create happens in the specific report like form3x). This is the same as if you hit a post /reports/something-that-does-not-exist/. That will also get routed to the create action and say 405.

.circleci/config.yml
```diff
@@ -55,7 +55,7 @@ jobs:
       - run:
           name: Run tests
           # Use built-in Django test module
-          command: coverage run --rcfile=.coveragerc manage.py test --parallel 4 --timing
+          command: coverage run --rcfile=.coveragerc manage.py test --parallel 4 --timing --settings=fecfiler.settings.test
```
toddlees: what do you think of using @patch("fecfiler.contacts.views.settings.E2E_TEST", True) in the reports test that failed like we did with the contacts endpoint?  I don't have strong feelings either way, but maybe whichever we pick we can tackle the issue the same way in both cases

toddlees (changes requested):

.circleci/config.yml
sasha-dresden: I found two ways to handle this. The first way keeps the e2e paths within the router for non-e2e servers, and instead just returns 404 not found, but technically the path does exist. The ticket mentioned trying to find a way to make it so it doesn't exist in the router at all, so I managed to find a way to completely remove it from the router when E2E_TEST is False. However, it does make testing very difficult, because it removes it a little too well haha. And what I mean is that it removes it during compile time, so patch doesn't work, because you can't patch back in what doesn't exist. However if you set the server as an E2E_TEST you can patch it out with False. It might make sense to run tests in two stages, one where you pretend to be an e2e enabled environment and one standard.

django-backend/fecfiler/reports/views.py
```diff
@@ -154,18 +155,20 @@ def unamend(self, request, pk):
...
-    )
-    def e2e_delete_all_reports(self, request):
-        reports = Report.objects.filter(committee_account__committee_id="C99999999")
-        report_count = reports.count()
+    if settings.E2E_TEST:
+
+        @action(
+            detail=False,
+            methods=["post"],
+            url_path="e2e-delete-all-repo",
```
toddlees: i think this is a breaking change

toddlees (changes requested):

django-backend/fecfiler/reports/tests/test_views.py
```diff
@@ -75,6 +76,38 @@ def test_ordering(self):
...
+        transaction_count = Report.objects.filter(
+            committee_account__committee_id="C99999999"
+        ).count()
+        self.assertGreater(report_count, 0)
+        self.assertGreater(transaction_count, 0)
+
+        response = self.client.post(
+            "/api/v1/reports/e2e-delete-all-reports/", data={}, format="json"
+        )
+        self.assertEqual(response.status_code, 403)
```
toddlees: you're getting a 403 here because the request is not authenticated rather than whether or not you're turning off the endpoint.  Try using the `self.send_viewset_post_request`

django-backend/fecfiler/contacts/tests/test_views.py
```diff
@@ -357,23 +356,21 @@ def test_e2e_delete_all_contacts(self, mock_settings):
...
         active_contact, deleted_contact = self._create_active_and_deleted_contacts(
             self.default_committee.id
         )
-        response = self._post_e2e_delete_all_contacts()
+        response = self.client.post(
+            "/api/v1/contacts/e2e-delete-all-contacts/", data={}, format="json"
+        )
 
-        self.assertEqual(response.status_code, 404)
+        self.assertEqual(response.status_code, 403)
```
toddlees: you're getting a 403 here because the request is not authenticated rather than whether or not you're turning off the endpoint.  Try using the `self.send_viewset_post_request`

toddlees (dismissed):

django-backend/fecfiler/reports/views.py
sasha-dresden: ouch, this is why we have code reviews. somehow managed to delete part of the uri.

django-backend/fecfiler/reports/tests/test_views.py
sasha-dresden: Ah, that's a good point. The reason I was not using `send_viewset_post_request` is because that required the action endpoint which wouldn't exist so it would throw a method doesn't exist error. However, I found a way to do it with authentication and now it properly returns 405.

django-backend/fecfiler/contacts/tests/test_views.py
sasha-dresden: Ah, that's a good point. The reason I was not using `send_viewset_post_request` is because that required the action endpoint which wouldn't exist so it would throw a method doesn't exist error. However, I found a way to do it with authentication and now it properly returns 405.

Elaine-Krauss-TCG (approved):

sasha-dresden:
As discussed in the dev sync up, I got rid of the test settings and instead just set E2E_TEST=True inside the circle config for running the e2e tests. Worked perfectly.

PR #2057, opened 2026-05-27
fosec-112 - add committee id to 'already exists' error logging

dheitzer:
Ticket link: https://fecgov.atlassian.net/browse/FOSEC-112
  
Related PRs:

### DIFF
--- a/django-backend/fecfiler/committee_accounts/views.py
+++ b/django-backend/fecfiler/committee_accounts/views.py
@@ -69,14 +69,21 @@ class CommitteeViewSet(viewsets.GenericViewSet, mixins.ListModelMixin):
 
     @action(detail=False, methods=["post"])
     def create_account(self, request):
-        committee_id = request.data.get("committee_id")
-        if not committee_id:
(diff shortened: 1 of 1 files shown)


Elaine-Krauss-TCG (approved):

danguyf (approved):

django-backend/fecfiler/committee_accounts/views.py
```diff
@@ -87,7 +94,7 @@ def get_available_committee(self, request):
             return Response(committee)
         except Exception as e:
             logger.error(
-                f"User {request.user.email} failed to create committee account "
+                f"User {request.user.email} failed to retrieve committee account "
```
toddlees: since we're using this endpoint during creation, and likely this will be the place that cuts them off, it might be still be good to say "create" here.  maybe "failed to retrieve committee for account creation?"

toddlees (changes requested):

toddlees (approved):

PR #2058, opened 2026-05-28
Patch fix for amending a report after the first amendment

toddlees:
Ticket link: https://fecgov.atlassian.net/browse/FECFILE-3148
  
Related PRs:  [App](https://github.com/fecgov/fecfile-web-app/pull/4000)

Addresses two main issues.  

1. One is the bug identified in the ticket.   The issue was that we were setting ther report id in the .fec to whatever the last report id was, so for amendment 1 the report id in the .fec was the one that came back with the new filing.  For the second and onward, EFO wants the report id to still be the original and not the new id of the previous amendment.  Now, when the response comes back from the EFO API, we only set the `fec_report_id` if it's the first filing, and not on amendments.  (note about `fec_report_id` below)

2. This came about when we fixed a previous bug https://fecgov.atlassian.net/browse/FECFILE-2853.  Fixing this bug broke that ticket's solution, so we had to come up with a different one.  The fix is to keep a key to the previous submission record so that if the user unamends, it'll just set the upload submission record to the previous.

I also deprecated report.report_id which i think is a candidate for most confusing variable name.  it's been replaced by `fec_report_id` to differentiate it with the uuid of the report in fecfile+.


django-backend/fecfiler/web_services/tests/test_models.py
```diff
@@ -64,6 +65,46 @@ def test_save_error(self):
         self.assertEqual(from_db.fecfile_task_state, FECSubmissionState.FAILED.value)
         self.assertIsNotNone(from_db.task_completed)
 
+    def test_amendment_save_fec_response_does_not_overwrite_report_id(self):
```
dheitzer, line 68: great test

dheitzer (approved):
looks great! Only thought would be - perhaps we want to add a ticket to remove the deprecated `report.report_id`

django-backend/fecfiler/reports/migrations/0009_rename_report_id_report_fec_report_id.py
```diff
@@ -0,0 +1,37 @@
...
+            name="previous_upload_submission",
+            field=models.ForeignKey(
+                blank=True,
+                null=True,
+                on_delete=models.deletion.SET_NULL,
+                related_name="reports_previous_upload_submission",
+                to="web_services.uploadsubmission",
+            ),
+        ),
+        migrations.RunPython(copy_report_id_to_fec_report_id),
```
Elaine-Krauss-TCG, line 36: we might want to specify the reverse code as `migrations.RunPython.noop`  for consistency's sake.  Not a blocker, though, in my opinion.

Elaine-Krauss-TCG (approved):

danguyf (approved):

sasha-dresden (approved):

PR #2071, opened 2026-06-02
feature/3158 - Remove Report.report_id

sasha-dresden:
Issue [FECFILE-3158](https://fecgov.atlassian.net/browse/FECFILE-3158)


toddlees (changes requested):
block until we have a deploy with the deprecation step.  will clarify

django-backend/fecfiler/reports/views.py
```diff
@@ -212,4 +212,4 @@ def filter_by_report(queryset, viewset):
         if viewset.request
         else None
     )
-    return queryset.filter(report_id=report_id) if report_id else queryset
+    return queryset.filter(fec_report_id=report_id) if report_id else queryset
```
toddlees, line 215: Haha my ide tried to get me to do this too.  this is actually not correct.  queryset in this case is for the view that uses the reportviewmixin.  so `report_id=report_id` here actually is the django shorthand for the report's uuid like `report.id`.

toddlees (changes requested):

django-backend/fecfiler/reports/views.py
toddlees: oh i see you undid the change in the next commit

toddlees (approved):

PR #2073, opened 2026-06-02
Feature/3165 can only unamend in progress reports

toddlees:
Ticket link: https://fecgov.atlassian.net/browse/FECFILE-3165
  
- Adds that a report must be in the status of In-progress in order to be unamended. 
- Also adds API enforcement of amend and unamend

I could have just added the in-progress condition to the action in app, but that seemed like a step in the wrong direction given that we have an api variable claiming to state whether or not the report can be unamended or not.  I took the approach of making the value coming out of the api as the final say as to whether something can be unamended.  The trouble i ran into is that we have the logic of what causes can_unamend spread all over the code, so it's not clear what the reasoning is.  I could have continued this pattern and flipped the flag when the report gets set to in progress but again, that seemed like a step in the wrong direction, further spreading the logic out amongst the codebase.   I went with an iteration towards what I think will be an ideal. 

I made an annotation to the report manager called `can_unamend_new`. This will one day become `can_unamend`.  It's based of conditions:  is it in progress AND can_unamend.  My goal is to in the future, cut out the old can_unamend and replace it with `version > 0` (is it an amendment) and `transactions_dirty` (something to track if transactions have been modified since it was clean). right now, `can_unamend` is doing the work of both of those and a little bit more that's handled by adding the `is in progress` check.  I will make a follow up ticket to do implement this plan


sasha-dresden (approved):
This does fix the problem. I was just thinking we'd set can_unamend to False when unamend was the report was submitted. But I think if we go forward with your follow up ticket that would be a lot more definitive (using a database structure) rather than as by-products of the submit/amend/unamend process.

toddlees:
> This does fix the problem. I was just thinking we'd set can_unamend to False when unamend was the report was submitted. But I think if we go forward with your follow up ticket that would be a lot more definitive (using a database structure) rather than as by-products of the submit/amend/unamend process.

I had the same thought.  As it is we still have logic living all over wherever we do things.  The follow up should resolve that

PR #2078, opened 2026-06-04
[Sprint 88] Feature/2819 dah patch idea

dheitzer:
Moves conditional e2e endpoints into urls.py (similar to mock_oidc_provider conditional endpoints) and updates tests to look for 404 when `settings.E2E_TEST` is False.  

**note** `settings.E2E_TEST` cannot be patched in the test to generate the 404 since urls.py is evaluated on api startup.  Therefore, we want the unit tests to run with E2E_TEST as false so we can confirm the absence of these endpoints.  Running the tests tests with the `--settings=fecfiler.settings.test` flag will now fail the unit tests accordingly.


toddlees:
All patch ideas resloved

PR #2085, opened 2026-06-05
Feature/2738 - SQL improvements to submit to fec and submit to webprint

sasha-dresden:
Issue [FECFILE-2738](https://fecgov.atlassian.net/browse/FECFILE-2738)

Slight improvements to submit to fec and submit to webprint queries. Only really reduces number of queries by one and reduces the return parameters of another query so pretty minor, all things said. Didn't seem to really effect time spent, but I suppose it technically is an improvement, so, yea.

Silk testing this is a bit of a pain. There really isn't an easy way to test celery tasks with silk. I was able to do some time diffs for total time spent but nothing related to number of queries or the really useful query breakdowns that silk gives. So what I did was made a synchronous version of webprint to test it out. You can see that version in the original commit, but I removed it for the final commit so it wouldn't pollute our actions. But if you wanted to play around with it, pull down that commit (or just copy/paste).

### DIFF
--- a/django-backend/fecfiler/web_services/tasks.py
+++ b/django-backend/fecfiler/web_services/tasks.py
@@ -121,16 +121,13 @@ def log_polling_notice(attempts):
     elif duration_in_minutes >= 1:
         duration_string = f"{duration_in_minutes} minutes(s)"
 
-    logger.info(
-        f"""Submission queued for processing.  Polling every {
+    logger.info(f"""Submission queued for processing.  Polling every {
             interval} seconds for {
-            MAX_ATTEMPTS} attempts over {duration_string}"""
-    )
+            MAX_ATTEMPTS} attempts over {duration_string}""")
 
 
 @shared_task
 def submit_to_fec(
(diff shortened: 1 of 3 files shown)


django-backend/fecfiler/web_services/tasks.py
danguyf, line 74: Validate my understanding: While the ticket calls out this bit specifically, there's no way to `select_related("dot_fec")` here because this is where the dot_fec is created.  Which is to say: your implementation is complete despite not altering the "example" given in the ticket.

danguyf (approved):

django-backend/fecfiler/web_services/tasks.py
sasha-dresden: the `select_related("dot_fec")` was added in submit_to_fec() and submit_to_webprint(). I actually used `select_related("dot_fec__report__committee_account")` to get the full depth of what's used in there.

Elaine-Krauss-TCG (approved):

PR #2089, opened 2026-06-10
3186 -  Update form3/form3x serializers to share update/save process. Add tests for Form3Serializer.

sasha-dresden:
Issue [FECFILE-3186](https://fecgov.atlassian.net/browse/FECFILE-3186)


Elaine-Krauss-TCG (changes requested):
Code looks great, but I found two bugs:

1) If you create an F3 or an F3X, and then you try to create another report of the other type with the same/overlapping coverage dates, then it doesn't save and doesn't show any error messages on the front-end.

2) If you create an F3 and an F3X and then try to update either to have overlapping coverage dates with the other, an error message pops up saying that "some transactions would now fall outside your coverage dates" even though no transactions exist.

sasha-dresden:
Closed cause I realize I created this branch as a hotfix instead of as a feature and when you change the branch name it automatically closes any PRs.

sasha-dresden:
> Code looks great, but I found two bugs:
> 
> 1. If you create an F3 or an F3X, and then you try to create another report of the other type with the same/overlapping coverage dates, then it doesn't save and doesn't show any error messages on the front-end.
> 2. If you create an F3 and an F3X and then try to update either to have overlapping coverage dates with the other, an error message pops up saying that "some transactions would now fall outside your coverage dates" even though no transactions exist.

Good catch. I think this was a pre-existing issue, but this definitely seems like the time to address it! This is actually fixed with some front end changes. See the new app [PR4047](https://github.com/fecgov/fecfile-web-app/pull/4047). Essentially it was only fetching the existing form dates for one type not both.

PR #2094, opened 2026-06-11
[Sprint-89]Feature/2157 - Add Update Report Version

sasha-dresden:
Issue [FECFILE-2157](https://fecgov.atlassian.net/browse/FECFILE-2157)

APP [PR4042](https://github.com/fecgov/fecfile-web-app/pull/4042)


Elaine-Krauss-TCG (approved):

django-backend/fecfiler/reports/views.py
```diff
@@ -174,6 +174,37 @@ def e2e_delete_all_reports(self, request):
...
+                    f"Changed version of report {report.id} "
+                    f"from {original_version} to {amendment}"
+                )
+            )
+
+            return Response(ReportSerializer(report).data, status=200)
+
+        except Exception as e:
+            return Response(
+                {"detail": f"An error occurred while updating the report: {str(e)}"},
```
toddlees: Let's not include the exception in the response.  We've generally taken a stance of not outputting error information in our responses.  The static message is good though.

django-backend/fecfiler/reports/views.py
```diff
@@ -174,6 +174,37 @@ def e2e_delete_all_reports(self, request):
...
+    def update_version_number(self, request, pk):
+        try:
+            report: Report = self.get_object()
+        except Exception:
+            return Response({"detail": "Report not found."}, status=404)
+
+        payload = request.data
+        original_version = report.report_version
+        amendment = payload.get("amendment")
+        e_filing_id = payload.get("eFilingId")
```
toddlees, line 187: we need to also set the 
`original_amendment_date` for form 24s and set it to None on going back to the original

django-backend/fecfiler/reports/tests/test_views.py
```diff
@@ -397,3 +397,81 @@ def test_unamend(self):
...
+        updated_report = Report.objects.get(id=report.id)
+        self.assertEqual(updated_report.report_version, "2")
+        self.assertEqual(updated_report.form_type, "F3XA")
+        self.assertEqual(updated_report.fec_report_id, "FEC-112233")
+        self.assertEqual(response.data["id"], str(report.id))
+
+    def test_update_version_number_to_original(self):
+        report = create_form3x(self.committee, "2026-01-01", "2026-02-01", {})
+        payload = {"amendment": "0", "eFilingId": ""}
+        response = self.send_viewset_post_request(
```
toddlees, line 439: might be worth bringing the state of the report to an amendment before calling this.  otherwise the assertions afterwards are already what they need to be

toddlees (changes requested):
A great start! A few updates.  One to make f24 work

django-backend/fecfiler/reports/views.py
sasha-dresden: Oh man, I can't believe I missed that. Got it in there now.

django-backend/fecfiler/reports/views.py
sasha-dresden: Oh, yea, I suppose we don't want to send that, I had been using it for testing in case things went wrong. Got it removed.

django-backend/fecfiler/reports/tests/test_views.py
sasha-dresden: Haha yea that's a fair point.  Got it set properly as amended before returning it to the original via the manual process.

danguyf (approved):

toddlees (approved):
This is really great!

PR #2097, opened 2026-06-17
Block deleting linked contacts via API

### LINKED ISSUE #1828: DELETE /api/v1/contacts/:id allows deletion when has_transaction_or_report=true but should be blocked with 4xx response
I have a spec in `e2e-extended/contacts/contacts.delete.cy.ts` currently being skipped (xit) because it forces an API DELETE request which returns with a 200, not the expected 4xx. I’ve included the spec below for context.

 
   /**
   * DELETE /api/v1/contacts/:id allows deletion when has_transaction_or_report=true
   * but should be blocked with 403 or 409 response
   * needs changes in backend to enforce properly
   */
  xit('Linked: API delete is rejected (403/409)', () =&gt; {
    const fetchLinkedContact = (attempt = 1) =&gt;
      waitForLinkedContactStatusFromApi(true, attempt, {
        pageSize: 10,
        logLabel: 'Linked contact not marked as linked yet',
        errorLabel: 'Setup error',
      });

    return fetchLinkedContact().then((linkedContact) =&gt;
      ContactsDeleteHelpers.requestWithCookies(
        'DELETE',
        `http://localhost:8080/api/v1/contacts/${linkedContact.id}/`,
        undefined,
        { failOnStatusCode: false },
      ).then((response) =&gt; {
        let bodyPreview = '';
        try {
          const serialized = JSON.stringify(response.body ?? null);
          bodyPreview = serialized ?? '';
        } catch (error) {
          bodyPreview = String(error);
        }
        Cypress.log({
          name: 'deleteContact api',
          message: `status ${response.status} body ${bodyPreview.slice(0, 500)}`,
        });
        expect(
          response.status,
          'DELETE should be blocked for linked contact (expected 403/409)',
        ).to.be.oneOf([403, 409]);
      }),
    );
  });
});
 

### QA Notes ###
null

### DEV Notes ###
null

### Design ###
null

 See full ticket and images here: [FECFILE-2750](https://fecgov.atlassian.net/browse/FECFILE-2750)

arpitjain099:
This blocks DELETE /api/v1/contacts/:id when the contact is linked to a transaction or Form 1M report reference. Instead of soft deleting, the endpoint now returns 409 with a clear error.

I also added tests for both paths:
- unlinked contact delete still returns 204
- linked contact delete now returns 409 and keeps the contact

Validation run:
- flake8 fecfiler/contacts/views.py fecfiler/contacts/tests/test_views.py

I could not run the Django unit tests in this environment because PostgreSQL is not running and Docker daemon access is unavailable.

Closes #1828

### DIFF
--- a/django-backend/fecfiler/contacts/tests/test_views.py
+++ b/django-backend/fecfiler/contacts/tests/test_views.py
@@ -8,6 +8,7 @@ from fecfiler.committee_accounts.models import CommitteeAccount
 from ..views import ContactViewSet, DeletedContactsViewSet
 from .utils import create_test_individual_contact
 from fecfiler.shared.viewset_test import FecfilerViewSetTest
+from fecfiler.transactions.tests.utils import create_schedule_a
 
 CANDIDATE_RESULTS = [
     {"name": "LNAME, FNAME I", "candidate_id": "P60012143", "office_sought": "P"},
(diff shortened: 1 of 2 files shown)


toddlees:
Thank you for your contribution @arpitjain099.  We do not currently have #1828 as prioritized work.  If we do, we will bring up this branch.  For now we will close the pr.

I'm curious, what led you to this ticket in particular?  Knowing that might help us improve the the contribution experience for developers like yourself

arpitjain099:
Thanks @toddlees, no problem closing it.

Honest answer to your question: I do open-source security work, mostly supply-chain and software-integrity, and a lot of it on federal and public-interest projects. So I go through active repos in that space looking for concrete, unassigned issues where I can actually write and test the fix. #1828 stood out because silently soft-deleting a contact that's still referenced by a transaction or a Form 1M report is a data-integrity gap, and it had a clear fix that was easy to cover with tests.

If it's useful, I'm glad to point my time at whatever actually helps you. Dependency, CI, and supply-chain hardening, or anything security-tagged especially, since that's my main area. If you can tell me where external help is welcome, I'll focus there instead of guessing from the open-issue list.

PR #2145, opened 2026-07-17
feature/3242 - Switch to allow blank instead of null as front-end sends empty strings for unfilled fields

sasha-dresden:
Issue [FECFILE-3242](https://fecgov.atlassian.net/browse/FECFILE-3242)

APP [PR4169](https://github.com/fecgov/fecfile-web-app/pull/4169)

### DIFF
--- a/django-backend/fecfiler/feedback/serializers.py
+++ b/django-backend/fecfiler/feedback/serializers.py
@@ -7,6 +7,6 @@ logger = structlog.getLogger(__name__)
 
 class FeedbackSerializer(serializers.Serializer):
     action = CharField(max_length=2000)
-    feedback = CharField(max_length=2000, allow_null=True)
-    about = CharField(max_length=2000, allow_null=True)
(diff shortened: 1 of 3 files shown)


danguyf (approved):

danguyf (changes requested):
This fixes the null vs blank issue but I just noticed the the DEV Notes also request a unit test for this case AND a solution to running it locally without the github error.

sasha-dresden:
> This fixes the null vs blank issue but I just noticed the the DEV Notes also request a unit test for this case AND a solution to running it locally without the github error.

Oh good catch, I forgot to check the dev notes :shame:. So I updated so it'll work when run in a "mock" state and also added a test here. I also updated the front-end to be a bit cleaner and with some tests there for handling invalid data.

danguyf (approved):
Awesome.

Elaine-Krauss-TCG (approved):

PR #2147, opened 2026-07-17
Feature/3233 - Unassociated transactions endpoint

Elaine-Krauss-TCG:
Ticket link: https://fecgov.atlassian.net/browse/FECFILE-3233
  
Related PRs:

### DIFF
--- a/django-backend/fecfiler/transactions/views.py
+++ b/django-backend/fecfiler/transactions/views.py
@@ -152,19 +152,17 @@ class TransactionViewSet(CommitteeOwnedViewMixin, ModelViewSet):
 
     @action(detail=False, methods=["get"], url_path=r"list/unassociated")
     def list_unassociated_transactions(self, request, *args, **kwargs):
+        if "page" not in request.query_params or request.query_params["page"] is None:
+            return Response("page is required", status=400)
+
         queryset = self.filter_queryset(self.get_queryset())
 
         # __isnull evaluates to true when no many-to-many relationships exist
         queryset = queryset.filter(reports__isnull=True)
 
-        if "page" in request.query_params:
-            page = self.paginate_queryset(queryset)
-            if page is not None:
-                serializer = self.get_serializer(page, many=True)
-                return self.get_paginated_response(serializer.data)
-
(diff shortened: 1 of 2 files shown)


django-backend/fecfiler/transactions/views.py
```diff
@@ -150,6 +150,22 @@ def get_queryset(self):
 
         return queryset
 
+    @action(detail=False, methods=["get"], url_path=r"list/unassociated")
+    def list_unassociated_transactions(self, request, *args, **kwargs):
+        queryset = self.filter_queryset(self.get_queryset())
+
+        # __isnull evaluates to true when no many-to-many relationships exist
```
dheitzer, line 160: great clarification

dheitzer (approved):

danguyf (approved):
Swagger does not surface the page parameter but I don't know if that's a concern.

django-backend/fecfiler/transactions/views.py
```diff
@@ -150,6 +150,22 @@ def get_queryset(self):
...
+        queryset = queryset.filter(reports__isnull=True)
+
+        if "page" in request.query_params:
```
toddlees: Let's make pagination required.  I noticed while reviewing another ticket that a lot of our list endpoints allow to optionally return everything unpaginated.  I made this ticket to address it https://fecgov.atlassian.net/browse/FOSEC-130.

toddlees (changes requested):
Nice and simple solution.  I was concerned that `report_type` would be an issue in `get_queryset` but i don't even think it's doing what it says it's doing what we think it's doing.  making a ticket for that

django-backend/fecfiler/transactions/views.py
Elaine-Krauss-TCG: Sounds good!

toddlees (changes requested):
comment

django-backend/fecfiler/transactions/views.py
Elaine-Krauss-TCG: Pagination is now required.  Submitting a request to the endpoint without the page query param will result in a 400 error.

PR #2203, opened 2026-08-13
FECFILE-3358/3359: fix boolean flag handling, update email error logging

danguyf:
Ticket link:
https://fecgov.atlassian.net/browse/FECFILE-3358
https://fecgov.atlassian.net/browse/FECFILE-3359

Related PRs:
N/A

### DIFF
--- a/django-backend/fecfiler/email.py
+++ b/django-backend/fecfiler/email.py
@@ -24,8 +24,8 @@ def send_email_notification(
 ):
     if SES_CLIENT is None:
         raise RuntimeError(
-            "Email is not enabled. To enable email, ensure FLAG__ENABLE_EMAIL is True "
-            "and SES_ACCESS_KEY_ID, SES_SECRET_ACCESS_KEY, and SES_REGION are set."
+            "SES client is not available. To enable email, ensure FLAG__ENABLE_EMAIL is "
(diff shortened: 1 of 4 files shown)


django-backend/fecfiler/email.py
```diff
@@ -24,7 +24,7 @@ def send_email_notification(
 ):
     if SES_CLIENT is None:
         raise RuntimeError(
-            "SES client is not available. Ensure FLAG__ENABLE_EMAIL is True "
+            "Email is not enabled. To enable email, ensure FLAG__ENABLE_EMAIL is True "
```
toddlees: This is less clear to me.  I think we're asking too much of `SES_CLIENT is NONE` to determine one error message.
The previous error makes sense to me.

Somewhere, (i realize you'd like it here and that's fine because i understand you want to keep it in one place) we should have something like 

```
if not EMAIL_ENABLED:
   logger.info("Not sending email because email is disabled")
   return

```
because it is not an error for us to have turned off email, where it is an error to have it turned on but not have the client configured correctly.

I still hold that we should be able to know if we intend to send an email before we call a function called "sendAddUserToCommitteeEmail" but like i said i wont die on that hill.

toddlees (changes requested):

django-backend/fecfiler/email.py
danguyf: Fiiiiiiiine.

toddlees (approved):

PR #2223, opened 2026-08-19
[Hotfix] Respond with 403 when no committee is activated

lbeaufort:
Ticket link: https://fecgov.atlassian.net/browse/FOSEC-139

I confirmed my unit test catches the server error in the prior commits.
  
Example request to local (log in first, don't select a committee)
```
curl -X 'GET' \
  'http://localhost:8080/api/v1/reports/' \
  -H 'accept: application/json'
 ```

  **Before:** 500 error
  **After:** 403

### DIFF
--- a/django-backend/fecfiler/shared/viewset_test.py
+++ b/django-backend/fecfiler/shared/viewset_test.py
@@ -164,6 +164,6 @@ class FecfilerViewSetTest(TestCase):
                 committee if committee is not None else self.default_committee
             )
             request.session = {
-                "committee_uuid": str(request_committee.id),
-                "committee_id": str(request_committee.committee_id),
+                "committee_uuid": str(request_committee.id) if request_committee else None,
+                "committee_id": str(request_committee.committee_id) if request_committee else None,
             }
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Review by toddlees — APPROVED** [review]
worked for me 

Image

File: django-backend/fecfiler/shared/viewset_test.py [hunk 1]
```diff
@@ -164,6 +164,10 @@ def init_request(self, request, authenticate=True, user=None, committee=None):
                 committee if committee is not None else self.default_committee
```
lbeaufort, line 164: It doesn't look like this is really handling `committee = None` when that's the intended scenario - for now, just overriding the default committee in my test with `super().set_default_committee(None)`, but this could probably use a refactor at some point

Elaine-Krauss-TCG (approved):

django-backend/fecfiler/shared/viewset_test.py
lbeaufort: @danguyf has a sample branch https://github.com/fecgov/fecfile-web-api/compare/develop...feature/hotfix-committee-none

PR #2226, opened 2026-08-19
Hotfix - Validates parent_transaction, loan, debt, reatt_redes

toddlees:
Ticket link:
  
Related PRs:


Elaine-Krauss-TCG:
Potential fix [PR](https://github.com/fecgov/fecfile-web-api/pull/2241) for breaking unit tests (also includes new test)

lbeaufort:
Bringing in @sasha-dresden's comments to consolidated PR https://github.com/fecgov/fecfile-web-api/pull/2241#pullrequestreview-4976899986

lbeaufort:
Capturing comment from @toddlees 

make sure it's comparing IDs - might be an id in one or the other or both - need to handle all scenarios

the only thing i was hoping to catch with that condition was that if there were two id's they must be the same.  I think there might be a flaw of that approach because i was really trying to just not do the lookup for both cases. Say there was no parent_transaction_id but there was the a bad id in the parent_transaction_object. Then the check as it's going to be is not sufficient. Outside or inside you shouldn't be able to make use of the ID

It might work to just make sure the IDs are the same and check the ID afterwards. So a combination of the "they must be the same, or there's no id at all", and "the single id must be your transaction"

django-backend/fecfiler/transactions/tests/test_views.py
```diff
@@ -2719,7 +2719,7 @@ def test_committee_locked_parent_transaction(self):
             other_committee.id
         )
         evil_contact = create_test_individual_contact(
-            "Good",
+            "Evil",
```
lbeaufort, line 2722: 😁

lbeaufort (approved):
Confirmed @toddlees approves of @Elaine-Krauss-TCG's changes

PR #2230, opened 2026-08-19
Hotfix/contact committee mixin

sasha-dresden:
Add contact committee validation mixin and use for both reports (F1M) and transactions


sasha-dresden:
Accidentally PR'd into develop instead of the unified

PR #2231, opened 2026-08-19
Hotfix - contact committee validation mixin

sasha-dresden:
Creates contact committee validation mixin and uses to validate contacts in report (F1M) and transactions.

### DIFF
--- a/django-backend/fecfiler/contacts/serializers.py
+++ b/django-backend/fecfiler/contacts/serializers.py
@@ -178,7 +178,7 @@ class ContactCommitteeValidationMixin:
             ).exists()
 
             if not exists:
-                errors[contact_field] = "Invalid contact id or contact does not belong."
+                errors[contact_field] = "Invalid contact id or contact."
 
         if errors:
             raise ValidationError(errors)
--- a/django-backend/fecfiler/contacts/serializers.py
+++ b/django-backend/fecfiler/contacts/serializers.py
@@ -149,32 +149,36 @@ class ContactCommitteeValidationMixin:
     """Mixin to validate that contact ids belongs to
     the serializer's committee_account."""
 
-    def validate_contact_committee_ownership(self, data, contact_field):
+    def validate_contact_committee_ownership(self, data, contact_fields):
+        if isinstance(contact_fields, str):
+            contact_fields = [contact_fields]
+
         committee_account = data.get("committee_account")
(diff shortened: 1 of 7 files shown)


django-backend/fecfiler/contacts/serializers.py
```diff
@@ -144,3 +143,38 @@ class Meta:
...
+        if contact_id:
+            if (
+                contact and contact.get("id") is not None
+                and contact.get("id") != contact_id
+            ):
+                raise ValidationError(
+                    {
+                        contact_field: (
+                            "Invalid contact id or contact does not belong "
+                            "to this committee account."
```
dheitzer: it seems like this error doesn't have anything to do with the committee account?

django-backend/fecfiler/reports/form_1m/serializers.py
```diff
@@ -183,6 +187,12 @@ def validate(self, data):
         self.context["fields_to_ignore"] = self.context.get(
             "fields_to_ignore", ["filer_committee_id_number"]
         )
+        self.validate_contact_committee_ownership(data, 'contact_affiliated')
```
dheitzer: would it be worth it to make the validation function take a list of contact data and use __in list (similar to report)?

django-backend/fecfiler/reports/serializers.py
```diff
@@ -225,7 +230,7 @@ def validate_report_committee_ownership(self, data):
 
         report_ids = data.get("report_ids", [])
         if report_ids:
-            unique_ids = set(str(rid) for rid in report_ids) if report_ids else set()
+            unique_ids = {str(rid) for rid in report_ids}
```
dheitzer, line 232: maybe we can take this approach to the contact validate function since we're often dealing with multiple contacts.

django-backend/fecfiler/transactions/serializers.py
```diff
@@ -281,6 +285,11 @@ def validate(self, data):
 
         self.validate_report_committee_ownership(initial_data)
         self.validate_memo_text_committee_ownership(data)
+        self.validate_contact_committee_ownership(data, 'contact_1')
```
dheitzer: same comment as above

dheitzer (changes requested):
just a couple thoughts

dheitzer (approved):

PR #2235, opened 2026-08-19
Hotfix - Unit test: transactions with another committee's child transactions

danguyf:
Unit tests (-- currently failing by design --) checking that create and update of transactions are not allowed to include another committee's child transactions.

I also got tired of logger choking on exceptions so I modified `custom_exception_handler` to log them without errors.

PR #2236, opened 2026-08-19
Hotfix - fix transaction children

danguyf:
Don't create or update a transaction with child transactions belonging to another committee.

After this change the new unit tests added in the related PR pass.

Related PRs:
https://github.com/fecgov/fecfile-web-api/pull/2235 (also included in this PR)

# Manual Reproduction

## In brief
* have a Committee A and a Committee B
* create a valid transaction under Committee B
* grab the child ID from that new transaction
* submit a new/create or update request under Committee A that includes that foreign child ID in `children`
* API should reject it with HTTP 400

### Prepare two committees and reports
In the app, or in test data, create:
* Committee A
* Committee B
* one report in Committee A
* one report in Committee B

### Create a valid child transaction in Committee B
* Open the Network tab of Developer Tools
* Create a normal transaction that has a child.
* Grab that API call as curl

### Create the malicious payload in Committee A
* Switch to Committee A and open the create transaction form.
* Maybe create a transaction there just to grab the API call and get the ID of Committee B
* Modify the API call to replace the `children` with the foreign child ID from Committee B

### Submit it
It should return an HTTP 400

### DIFF
--- a/django-backend/fecfiler/transactions/views.py
+++ b/django-backend/fecfiler/transactions/views.py
@@ -484,7 +484,7 @@ class TransactionViewSet(CommitteeOwnedViewMixin, ModelViewSet):
         ).filter(id=child_id, committee_account_id=committee_id).first()
         if child_instance is None:
             raise ValidationError(
-                {"children": ["Invalid child_id or child does not belong."]}
+                {"children": ["Invalid child_id or child"]}
             )
         return child_instance
 
--- a/django-backend/fecfiler/transactions/views.py
+++ b/django-backend/fecfiler/transactions/views.py
@@ -5,6 +5,7 @@ from rest_framework.filters import OrderingFilter
 from rest_framework.decorators import action
 from rest_framework.response import Response
 from rest_framework import status
+from rest_framework.exceptions import ValidationError
 from rest_framework.viewsets import ModelViewSet
 from datetime import datetime
 from django.db.models import Q
@@ -472,6 +473,17 @@ class TransactionViewSet(CommitteeOwnedViewMixin, ModelViewSet):
             ):
                 target.schedule_f.aggregate_general_elec_expended -= amount
 
+    def get_child_instance(self, child_id, committee_id):
+        # ensure that child transaction belongs to this committee
+        child_instance = Transaction.objects.select_related(
+            "schedule_a", "schedule_b", "schedule_e", "contact_2"
+        ).filter(id=child_id, committee_account_id=committee_id).first()
+        if child_instance is None:
+            raise ValidationError(
+                {"children": ["Child transaction does not belong to this committee."]}
+            )
+        return child_instance
+
     def save_transaction(self, transaction_data, request):
         committee_id = request.session["committee_uuid"]
         children = transaction_data.pop("children", [])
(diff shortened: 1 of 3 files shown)


django-backend/fecfiler/transactions/views.py
```diff
@@ -591,39 +603,40 @@ def save_transaction(self, transaction_data, request):
...
+                    Schedule.B,
+                    Schedule.E,
+                ]:
+                    child_instances_to_aggregate.append(child_instance)
             else:
+                child_id = child_transaction_data.get("id")
+                if child_id:
+                    # we don't need the instance but...
+                    # we do want to check that it belongs to this committee
+                    child_instance = self.get_child_instance(child_id, committee_id)
```
Elaine-Krauss-TCG: We could just call the function without assigning it to a variable to avoid the style issue, but this isn't a blocking issue IMO

Elaine-Krauss-TCG (approved):

django-backend/fecfiler/transactions/views.py
danguyf: Good call!

django-backend/fecfiler/transactions/views.py
```diff
@@ -472,6 +473,17 @@ def adjust_running_totals(self, target: Transaction, source: Transaction):
...
                 target.schedule_f.aggregate_general_elec_expended -= amount
 
+    def get_child_instance(self, child_id, committee_id):
+        # ensure that child transaction belongs to this committee
+        child_instance = Transaction.objects.select_related(
+            "schedule_a", "schedule_b", "schedule_e", "contact_2"
+        ).filter(id=child_id, committee_account_id=committee_id).first()
+        if child_instance is None:
+            raise ValidationError(
+                {"children": ["Child transaction does not belong to this committee."]}
```
dheitzer: per team discussion we don't want to reveal this information in our error messages.

django-backend/fecfiler/utils.py
```diff
@@ -7,11 +9,32 @@
...
+            except TypeError:
+                return str(detail)
+        elif getattr(value, "args", None):
+            value = value.args[0] if len(value.args) == 1 else value.args
+
+    if isinstance(value, (dict, list, tuple, set)):
+        try:
+            return json.dumps(value, default=str, sort_keys=True)
+        except TypeError:
+            pass
```
dheitzer, line 30: do we want to return value here?

django-backend/fecfiler/transactions/views.py
danguyf: How about now?

django-backend/fecfiler/utils.py
danguyf: We do, the next line returns. Or should I have `noop`?

dheitzer: yep nevermind :) pr indentation got me

django-backend/fecfiler/transactions/views.py
danguyf: Todd had me revise further.

toddlees (approved):

PR #2246, opened 2026-08-20
Hotfix - unit tests for creating/updating transactions with loan from another committee

danguyf:
The fix for the issue is already in place, now we have unit tests as well to guard against future reversions.


toddlees (approved):

lbeaufort (approved):
Awesome!

PR #2261, opened 2026-08-25
fosec-133 process log tokens

dheitzer:
Ticket link: https://fecgov.atlassian.net/browse/FOSEC-133
  
Related PRs:

### DIFF
--- a/django-backend/fecfiler/settings/base.py
+++ b/django-backend/fecfiler/settings/base.py
@@ -439,8 +439,10 @@ def add_migration_logs(logger: logging.Logger, method_name: str, event_dict):
 
 
 def process_log_tokens(logger, method_name, event_dict):
-    event_dict["event"] = re.sub(r'user ".*?"', 'user ""', event_dict["event"])
-    event_dict["event"] = re.sub(r'database ".*?"', 'database ""', event_dict["event"])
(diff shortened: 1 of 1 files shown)


django-backend/fecfiler/settings/base.py
```diff
@@ -404,6 +438,12 @@ def add_migration_logs(logger: logging.Logger, method_name: str, event_dict):
     return event_dict
 
 
+def process_log_tokens(logger, method_name, event_dict):
+    event_dict["event"] = re.sub(r'user ".*?"', 'user ""', event_dict["event"])
+    event_dict["event"] = re.sub(r'database ".*?"', 'database ""', event_dict["event"])
```
danguyf: Should we consider replacing with `[redacted]` rather than an empty string to differentiate between logging errors where the user or database wasn't passed and ones where we're redacting it?

toddlees (approved):

danguyf (approved):
I learned things!

django-backend/fecfiler/settings/base.py
dheitzer: Good though @danguyf.  I had considered whether this would draw attention to the potential issue.  Updated to "****".

PR #2269, opened 2026-08-27
Moving gunicorn env variables to be in defaults rather than ony api

toddlees:
Ticket link:
  
Related PRs:


dheitzer (approved):

lbeaufort (approved):
Nice, thanks! Only changing manifests and the develop build is broken, so merging before unit tests finish.

PR #2292, opened 2026-09-09
FECFILE-2421: delete load test mirror data

danguyf:
Ticket link:
https://fecgov.atlassian.net/browse/FECFILE-2421

Related PRs:
N/A

* unit tests pass
* successfully tested in load mirror by running `gen_locust_load_test_data`, `get_overview` to get the counts, then `delete_locust_load_test_data`, then `get_overview` again

### DIFF
--- a/django-backend/fecfiler/devops/management/commands/delete_locust_load_test_data.py
+++ b/django-backend/fecfiler/devops/management/commands/delete_locust_load_test_data.py
@@ -5,7 +5,7 @@ from fecfiler.devops.utils.load_test import LoadTestUtils
 
 
 class Command(FECCommand):
-    help = "Delete Locust load test data"
+    help = "Delete locust test data from load mirror"
     command_name = "delete_locust_load_test_data"
 
     def command(self, *args, **options):
--- a/django-backend/fecfiler/devops/management/commands/delete_locust_load_test_data.py
+++ b/django-backend/fecfiler/devops/management/commands/delete_locust_load_test_data.py
@@ -1,0 +1,18 @@
+from django.core.management.base import CommandError
+
+from .fecfile_base import FECCommand
+from fecfiler.devops.utils.load_test import LoadTestUtils
+
+
+class Command(FECCommand):
+    help = "Delete Locust load test data"
+    command_name = "delete_locust_load_test_data"
+
+    def command(self, *args, **options):
+        load_test_utils = LoadTestUtils()
+        try:
+            load_test_utils.validate_load_mirror_runtime()
+        except ValueError as error:
+            raise CommandError(str(error)) from error
+
+        load_test_utils.delete_load_test_committees_and_data()
--- a/django-backend/fecfiler/devops/tests/test_load_test_utils.py
+++ b/django-backend/fecfiler/devops/tests/test_load_test_utils.py
@@ -135,9 +135,6 @@ class LoadTestUtilsTestCase(TestCase):
                     "aws-rds": [
                         {"name": "fecfile-api-rds"},
                     ],
-                    "s3": [
-                        {"name": "load-fecfile-api-s3"},
-                    ],
                 }
             ),
         },
--- a/django-backend/fecfiler/devops/tests/test_load_test_utils.py
+++ b/django-backend/fecfiler/devops/tests/test_load_test_utils.py
@@ -1,6 +1,9 @@
 from django.test import TestCase
 from unittest.mock import patch, MagicMock
 from fecfiler.devops.utils.load_test import LoadTestUtils
+from fecfiler.committee_accounts.models import CommitteeAccount, Membership
+from fecfiler.user.models import User
+import json
 
 
 class LoadTestUtilsTestCase(TestCase):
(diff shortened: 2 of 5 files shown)


django-backend/manage.py
```diff
@@ -38,6 +38,7 @@
     # COMMITTEE ACCOUNT COMMANDS #
     "load_mocked_committee_data",
     "gen_locust_load_test_data",
+    "delete_locust_load_test_data",
```
lbeaufort, line 41: Agree that putting this in `restricted_commands` makes sense as an additional protection

django-backend/fecfiler/devops/tests/test_load_test_utils.py
```diff
@@ -69,3 +72,176 @@ def test_create_new_committee(
...
+                    ],
+                    "s3": [
+                        {"name": "load-fecfile-api-s3"},
+                    ],
+                }
+            ),
+        },
+        clear=False,
+    )
+    def test_validate_load_mirror_runtime_fails_for_non_load_app(self):
```
lbeaufort, line 121: neat test!

django-backend/fecfiler/devops/management/commands/delete_locust_load_test_data.py
```diff
@@ -0,0 +1,18 @@
...
+    command_name = "delete_locust_load_test_data"
+
+    def command(self, *args, **options):
+        load_test_utils = LoadTestUtils()
+        try:
+            load_test_utils.validate_load_mirror_runtime()
+        except ValueError as error:
+            raise CommandError(str(error)) from error
+
+        load_test_utils.delete_load_test_committees_and_data()
```
dheitzer: any reason not to put this inside the try?

django-backend/fecfiler/devops/utils/load_test.py
```diff
@@ -8,9 +12,120 @@
...
+REQUIRED_LOAD_RDS_SERVICE = "load-fecfile-api-rds"
+LOAD_TEST_CONTACT_CITY = "Testville"
+LOAD_TEST_CONTACT_EMPLOYER = "Business Inc."
+LOAD_TEST_CONTACT_OCCUPATION = "Job"
+LOAD_TEST_REPORT_FORM_TYPE = "F3XN"
 
 
 class LoadTestUtils:
+    def validate_load_mirror_runtime(self):
+        app_name, service_names = self.get_cloud_foundry_runtime_identity()
```
dheitzer, line 25: I wonder if we want to set an environment variable for only the load test api deployments and check that instead of needing to infer it from the app/services names, etc.  This would remove the need for the code to interface with cloud foundry and verifying the names and also require that we intentionally set the flag indicating a load test environment to be able to delete data on the instance.  Just a thought.

lbeaufort: @dheitzer interesting idea! We could probably make a separate manifest for the mirror site that has that env var set

django-backend/fecfiler/devops/utils/load_test.py
```diff
@@ -8,9 +12,120 @@
...
+        for service_instances in vcap_services_json.values():
+            for service_instance in service_instances:
+                service_name = service_instance.get("name")
+                if service_name:
+                    service_names.add(service_name)
+
+        return service_names
+
+    def delete_load_test_committees_and_data(self):
+        committees_to_delete = CommitteeAccount.all_objects.filter(
```
dheitzer, line 91: I'm wondering if we can be more intentional about which committees to delete by passing it into the command? In the generate load test data command, we are starting with a [base committee number](https://github.com/fecgov/fecfile-web-api/blob/3ce58f98ae8dc61b995b5e96e78ab5f019698fdf/django-backend/fecfiler/devops/management/commands/gen_locust_load_test_data.py#L10) to sequentially generate load test committees.  Maybe a similar approach for deleting them? Just an idea.

django-backend/fecfiler/devops/utils/load_test.py
```diff
@@ -8,9 +12,120 @@
 logger = structlog.get_logger(__name__)
 
 TEST_USER_EMAIL = "test@test.com"
+LOAD_PREFIX = "load-"
+REQUIRED_LOAD_RDS_SERVICE = "load-fecfile-api-rds"
+LOAD_TEST_CONTACT_CITY = "Testville"
```
dheitzer, line 17: My only concern is that if we ever change our load test data mocking or add to it, we will miss deleting those.  I'm wondering if we can be more explicit generally about the data we want to delete (comment added below).

dheitzer (dismissed):
just some thoughts

django-backend/fecfiler/devops/utils/load_test.py
```diff
@@ -8,9 +12,120 @@
...
+        service_names = set()
+        for service_instances in vcap_services_json.values():
+            for service_instance in service_instances:
+                service_name = service_instance.get("name")
+                if service_name:
+                    service_names.add(service_name)
+
+        return service_names
+
+    def delete_load_test_committees_and_data(self):
```
dheitzer, line 90: Great idea to be conservative/multiple checks about deleting data

django-backend/fecfiler/devops/utils/load_test.py
danguyf: I like the idea of a separate manifest and a feature flag, but I would want to keep some of those checks in there just in case someone mistakenly bound the live RDS service to the load mirror API app, for example.

django-backend/fecfiler/devops/utils/load_test.py
danguyf: I'd definitely rather that the fail case is *not* deleting data.

django-backend/fecfiler/devops/management/commands/delete_locust_load_test_data.py
danguyf: `validate_load_mirror_runtime()` is in the `try` because I want to test that we're in the load mirror and capture/raise it as a sort of pre-flight error if we aren't, whereas if the actual delete fails I'm fine with it stacktracing all over the terminal.

Including it in the `try` maybe makes a lot of sense, though...

django-backend/fecfiler/devops/utils/load_test.py
danguyf: I could see it. I'd want to add an arg so you could pass in a starting number or a range, since the gen_data command lets you pass in a starting committee ID.

lbeaufort (approved):
Looks great, thanks for including tests!

django-backend/fecfiler/devops/utils/load_test.py
danguyf: After sleeping on this, I can't come up with a use case that I like enough to implement this now. I don't want someone to have the ability to delete a range of committee IDs without other checks, so this would at most be an additional filter, meaning it could only be used to selectively delete only some of the test data.  There could be a call to allow that but I think it's beyond this ticket.

django-backend/fecfiler/devops/utils/load_test.py
danguyf: I suggest that we consider a follow-up task to create a separate manifest for the mirror site, after which we could re-consider putting this behind a feature flag.

django-backend/fecfiler/devops/management/commands/delete_locust_load_test_data.py
danguyf: I added a second try because I want to differentiate between the two sources of exceptions and handle them differently. Thoughts?

toddlees (approved):

PR #2305, opened 2026-09-15
[HOLD for revert of FECFILE-3257]Unassigned transactions and assigned transactions use the same list endpoint

sasha-dresden:
Issue [FECFILE-1819](https://fecgov.atlassian.net/browse/FECFILE-1819)

APP [PR4375](https://github.com/fecgov/fecfile-web-app/pull/4375)


danguyf (approved):
Slick reduction!

sasha-dresden:
After talking with @toddlees about this we decided to pass report_id=null so that it was clear that this was looking for ones without a report and so that in the future when we do the ticket to get all transactions you'll just not pass a report so it would be unlimited by reports.

toddlees (approved):

lbeaufort (changes requested):
https://fecgov.atlassian.net/browse/FECFILE-3257 is in the sprint backlog, want to make sure this doesn't get merged

PR #2307, opened 2026-09-16
bug/3384 - Form 3X entity aggregate totals are incorrect for unaggregated transactions

dheitzer:
Ticket link: https://fecgov.atlassian.net/browse/FECFILE-3384
  
Related PRs:

### About the diff
Black applied a lot of formatting changes to `utils_aggregation_service.py` but the only actual changes were the removal of the `force_unaggregated__isnull=True,` parameter on lines 117, 179, and 309

### Issues
Two code issues were fixed for this bug:

1. The `force_unaggregated__isnull=True,` filter was being added to places where we are retrieving transaction list to calculate the aggregates for.  Because of this, once force aggregate/unaggregate was applied, those transactions were not considered for aggregate recalculation.  This flag was removed.
2. Our `ENTITY_AGGREGATE_CLAUSE` was always using `effective_amount` as a base value to calculate aggregates, and thus was not taking into account the case where `force_unaggregated` flag is set.  This was fixed to match the behavior in `ELECTION_AGGREGATE_CLAUSE` which sets the aggregate to zero when this flag is set.

### Testing
These changes affect both entity and election aggregate calculations in the system.  Unit test was added and e2e tests were verified, however the following spots that should be checked for regression include:

- aggregate calculations for create/update transaction schedule A,B,E and child transactions
- transaction schedule A,B,E contact update (old contact and new contact aggregates should be updated correctly) 
- aggregate calculations for force aggregate/unaggregate for Schedule A,B,E
- aggregate calculations after transaction deleted from report
- aggregate calculations after report deletion

[e2e-smoke passing](https://app.circleci.com/pipelines/github/fecgov/fecfile-web-app/13438/workflows/357633ef-ae6a-4764-bf42-dfcfecd05d01/jobs/43901) 
[e2e-extended passing with 2 unrelated failures](https://app.circleci.com/pipelines/github/fecgov/fecfile-web-app/13438/workflows/41351739-be30-42e8-8251-1964abf11aab/jobs/43902)


django-backend/fecfiler/transactions/utils_aggregation_service.py
```diff
@@ -109,15 +107,17 @@ def calculate_entity_aggregates(
     """
     # Get all matching transactions ordered by date and created
     # Exclude force_unaggregated transactions as they don't get aggregated
```
sasha-dresden, line 109: So I noticed most of these fixes were simply removing the force unaggregated is null check, but I'm wondering if there was a reason for them based on this comment? Or maybe this comment is in error based on the bug we're having.

dheitzer: @sasha-dresden great catch.  This comment is in error I believe since while they don't get included in the aggregate, they do need to have their aggregate value recalculated when this flag is set.  This ties into the other change I made [here](https://github.com/fecgov/fecfile-web-api/pull/2307/changes#diff-909ced4f33e29e06f14baa9e8c9cc03b9c96f534a9b023648c2a4463a56bc986) to set the aggregate to 0 when it is set to True. 

My guess is we used to ignore records and just use effective_amount, but at some point that changed to us using 0 if the flag is present (`ELECTION_AGGREGATE_CLAUSE` for example), and these flags were not removed from the queries.

I'll remove the comments unless anyone more familiar with the aggregate work disagrees with my assessment.

sasha-dresden (approved):

django-backend/fecfiler/transactions/utils_aggregation_service.py
sasha-dresden: Yea, testing it, your changes work perfectly for the instructions in the ticket, so that sounds good to remove the comment. But yea, everything looks good to me.

dheitzer: I'm wondering if @toddlees might have more insight as it looks like the change to use effective_amount in the ENTITY_AGGREGATE_CLAUSE was added [here](https://github.com/fecgov/fecfile-web-api/pull/1869/changes/527d93827c50f59b05433c23e7cba8fbda953ac1#diff-909ced4f33e29e06f14baa9e8c9cc03b9c96f534a9b023648c2a4463a56bc986).  That change would seem to require that we include the force_unaggregate__null filter for retrieving transactions, but this also results in the transactions not having their aggregates recalculated when the flag is set and hence this bug.

toddlees (changes requested):
We're gonna need a migration, per the dev notes to update existing values

lbeaufort (commented):
Per @toddlees's comment, we ran into a performance issue with triggering aggregates in the past, and this PR seemed to address it: https://github.com/fecgov/fecfile-web-api/pull/1451/changes

There are also merge conflicts that need to be addressed

django-backend/fecfiler/transactions/tests/test_manager.py
```diff
@@ -89,7 +89,7 @@ def test_force_unaggregated(self):
         )
         for t in transactions:
             t.refresh_from_db()
-        self.assertEqual(transactions[0].aggregate, None)
+        self.assertEqual(transactions[0].aggregate, Decimal("0.00"))
```
toddlees, line 92: If i'm reading this test case correctly as checking the aggregate for the unaggregated transaction, here, we do want the aggregate to be None rather than $0

toddlees (changes requested):

PR #2317, opened 2026-09-21
FECFILE-2778: Log dumping committee data.

danguyf:
## Summary 

Ticket link:
https://fecgov.atlassian.net/browse/FECFILE-2778
  
We do not currently log when we run `dump_committee_data`, the closest we get is logging when it’s saved to S3 or local (but not redis).  So the first step is adding that logging.

Related PRs:
N/A

## Impacted areas of the application

Not user-facing. Only the logging of `dump_committee_data` is affected.

## How to test

There will be a related cloud.gov dashboard that queries the logs and shows any data dumping.  Testing could be accomplished by doing a data dump in a lower environment and then seeing if the dashboard displays it.


lbeaufort (commented):
Thanks for making this change! Do you have a sense of how easy it would be to see which user ran this management command in the cloud.gov logs? It might be worth making sure we can monitor who took the action.

toddlees (approved):

dheitzer (approved):

toddlees:
> Thanks for making this change! Do you have a sense of how easy it would be to see which user ran this management command in the cloud.gov logs? It might be worth making sure we can monitor who took the action.

@lbeaufort you can see who ran a command in `cf events fecfile-web-api`

